PR #48, opened 2016-09-28
database update improvements

Pseudodude:
- Avoid commits during update to improve performance
- Update from AllCards.json.zip instead of plaintext AllCards.json

### DIFF
--- a/database.py
+++ b/database.py
@@ -29,11 +29,13 @@ class Database():
     return r
 
   def value(self, sql, args = [], default = None):
-    rs = self.execute(sql, args)
-    try:
-      return rs[0][0]
-    except IndexError:
+    rs = self.database.execute(sql, args).fetchone()
+    if rs is None:
       return default
+    elif len(rs) <= 0:
+      return default
+    else:
+      return rs[0]
 
   def setup(self):
     self.execute("CREATE TABLE version (version TEXT)")
@@ -99,5 +101,5 @@ class Database():
       ('Special')
     """)
 
-# Import last to work around circular dependency â€” http://effbot.org/zone/import-confusion.htm
+# Import last to work around circular dependency — http://effbot.org/zone/import-confusion.htm
 import oracle
--- a/fetcher.py
+++ b/fetcher.py
@@ -1,4 +1,4 @@
-import json, urllib.request
+import json, os, shutil, urllib.request, zipfile
 
 from pkg_resources import parse_version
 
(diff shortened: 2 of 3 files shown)


database.py
```diff
@@ -29,11 +29,13 @@ def execute(self, sql, args = []):
     return r
 
   def value(self, sql, args = [], default = None):
-    rs = self.execute(sql, args)
-    try:
-      return rs[0][0]
-    except IndexError:
+    rs = self.database.execute(sql, args).fetchone()
```
bakert, line 32: rs is short for "resultset" or "recordset". If we're only fetching a single row then 'row' would be a more appropriate name.

database.py
```diff
@@ -29,11 +29,13 @@ def execute(self, sql, args = []):
...
+    if rs is None:
+      return default
+    elif len(rs) <= 0:
       return default
```
bakert, line 36: if rs is None or len(rs) == 0:
  return default

?

fetcher.py
```diff
@@ -10,7 +10,17 @@ def version(self):
     return parse_version(json.loads(self.open('https://mtgjson.com/json/version.json')))
 
   def all_cards(self):
-    return json.loads(self.open('https://mtgjson.com/json/AllCards.json'))
+    if os.path.isdir('./ziptemp'):
```
bakert, line 13: We let the user choose locations for db and images. I wonder if we should just have single entry in config.json for "directory to store stuff in" and put this in it too.

Then we can not delete it afterwards and check its timestamp before bothering to re-download it (if it's less than hour old, say).

fetcher.py
```diff
@@ -10,7 +10,17 @@ def version(self):
...
+      shutil.rmtree('./ziptemp')
+      
+    os.mkdir('./ziptemp')
+    urllib.request.urlretrieve('https://mtgjson.com/json/AllCards.json.zip', './ziptemp/AllCards.json.zip')
```
bakert, line 17: We should really make "download {url} to {location}" a method of fetcher and call it when fetching images from bluebones or gatherer, too.

oracle.py
```diff
@@ -73,20 +75,22 @@ def insert_card(self, name, card):
...
     sql += ')'
     values = [card.get(self.underscore2camel(property)) for property in Oracle.properties()]
-    self.database.execute(sql, values)
+    # self.database.execute commits after each statement, which we want to 
+    # avoid while inserting cards
+    self.database.database.execute(sql, values)
     id = self.database.value('SELECT last_insert_rowid()')
     for name in card.get('names', []):
-      self.database.execute('INSERT INTO card_name (card_id, name) VALUES (?, ?)', [id, name])
+      self.database.database.execute('INSERT INTO card_name (card_id, name) VALUES (?, ?)', [id, name])
```
bakert, line 83: Good God do I hate self.database.database :)

I feel like we're reaching into the internals of Database here. It might be nicer to build up an array of arrays of args and then process them all at the end? sqlite3 exposes executeMany which may or may not be useful.

If that's too much surgery then perhaps we can just expose a version of execute that doesn't commit or have a boolean param on execute for commit-or-not?

database.py
```diff
@@ -99,5 +101,5 @@ def setup(self):
       ('Special')
     """)
 
-# Import last to work around circular dependency — http://effbot.org/zone/import-confusion.htm
+# Import last to work around circular dependency � http://effbot.org/zone/import-confusion.htm
```
bakert, line 104: What changed here? Trailing whitespace? In a good way or a bad way?

bakert (approved):
Some things to think about but nothing blocking.

silasary:
Well, Jenkins is much happier with this one. Only took a minute to build, rather than the Hour that we used to deal with when building a new db.

oracle.py
Pseudodude: self.database.database could become self.database.connection, but I take your point about calling the function on Database's instance variable. I didn't want to change or rename the current execute+commit of my own accord, but adding a boolean param sounds like the best solution.

Issue #47, opened 2016-09-28
Support :10:, :UU:, :2W: emojis


bakert:
@binarySwordsman has added :00:, :10:, :11:, :12:, :13:, :15:, :16:, :01:, :2B:, :2G:, :2R:, :2U:, :2W:, :02:, :03:, :04:, :05:, :06:, :07:, :08:, :09:, :BG:, :BP:, :BR:, :BB:, :CC:, :EE:, :GP:, :GU:, :GW:, :GG:, :RG:, :RP:, :RW:, :RR:, :SS:, :TT:, :UB:, :UP:, :UR:, :UU:, :WB:, :WP:, :WU:, :WW:, :XX:

The rule is "remove curly braces and slashes. If single letter, double up. If single number prepend with 0."

Note there is currently no emoji for {E} (Energy), {Y}, {Z}, {Q} (untap), {P} (generic Phyrexian mana, only on one card's rules text), {20} (only in one card's rules text), nor any of the un-set specific symbols - {100}, {1,000,000}, {∞}, {hr}, {½}, {hw}.

bakert:
So we found out an interesting thing.

If the bot says ":UU:" we get that, literally.

What the bot actually has to say is: "<:UU:8392839283929>" (fake number but you get the idea).

The number is the image id of the uploaded image.

We could hardcode these values but I think a better approach is as follows:
- On startup the bot scans mana_cost and text columns for `/\{[^\}]*\}/` for all cards and uniqifies down to the 50 or so symbols used.
- It then sends itself a message with all the symbols in in some format: "!symbols {W} | {U} | {B} | ..."
- Upon receiving that message it sets up an in-memory dict of symbol => code which can be used to transform messages.
- We'll have to remove the check to make sure it doesn't respond to itself to make it respond to !symbols. It should _only_ respond to itself. The conversation should happen in a private message or maybe in #tech. Better if it's server-portable.

I'm going to bed now not sure if @binarySwordsman is implementing.

PR #51, opened 2016-09-28
Use emoji for mana costs

silasary:


silasary:
This closes #47

silasary:
Leaving this open for either someone else to look over, or for me to check in the morning.

bakert (approved):
Looks good to me. 

If it converts {T} to :TT: then "symbols2emojis" or some other name might be more appropriate as it's not exclusive to mana costs (even though that's the only way we're using it now).

BTW @binarySwordsman didn't actually add :QQ: due to a max of 50 emjois.

bot.py
```diff
@@ -48,6 +48,13 @@ def basename(cards):
 def unaccent(s):
   return ''.join((c for c in unicodedata.normalize('NFD', s) if unicodedata.category(c) != 'Mn'))
 
+def prettify_mana_cost(cost):
+  cost = re.sub(r'\{([0-9])\}', r':0\g<1>:', cost)
```
bakert, line 52: Not sure what this \g stuff is but:

```
re.sub(r'(a..)', r'*\1*', 'abcde abc xyzabc') # => '*abc*de *abc* xyz*abc*'
```

does what I would expect so you can maybe make that a little less verbose. Not important.

bakert (commented):
Also, @binarySwordsman did this another way because he didn't realize you had and merged to master. Also it doesn't seem to work (his way or your way) because messages from bots don't seem to respect custom emoji in the UI.

I've actually seen this with non-custom emoji on the mobile client (iOS). Not sure if that's related. I sometimes (always?) see :legal_checkmark_white: or whatever it is instead of the symbol.

He's reporting it to discord.

bot.py
silasary: Because the backref is immediately followed by a colon, it was complaining about an Invalid Group Reference.
The verbose \g<1> fixed that.

PR #59, opened 2016-09-29
database setup tweaks

silasary:


database.py
```diff
@@ -2,6 +2,9 @@
 import config
 
 class Database():
+  # Bump this if you modify the schema.
+  schema_version = 1
```
bakert, line 6: Oooooh I like this.

silasary: So much better than "delete `db` before you pull"

PR #60, opened 2016-09-29
Fix the case where the server has no emjois.

bakert:


silasary:
Ah. I knew to check that it wasn't a PM, but I had assumed a server with no emojis would just return `[]` for that property.

PR #64, opened 2016-09-29
Convert from 2 spaces to 4 as per https://www.python.org/dev/peps/pep-0008/

bakert:


silasary:
That is a horrifying diff

PR #73, opened 2016-09-29
Jenkins linting

silasary:


silasary:
For now, this is entirely diagnostic.

Should we desire, we can tell Jenkins to fail a build with too many warnings

PR #88, opened 2016-10-01
Download and store set and printing information in db.

bakert:
Sets are MtG sets. Pretty straightforward. They go in the table `set`.
You'll always want backticks on that because SET is an SQL keyword but
they are so very definitely called sets that I used it anyway.

Printings are individual printings of cards. Effectively card+set.

I moved the properties that are not populated in AllCards.json (on a 'card')
but are populated in AllSets.json (on a 'printing') to printing.

I changed the teardown of the database from a bunch of DROP statements to
a simple delete of the file. That way you don't have to remember to add
a table to the list. That said if we ever want persistent data we'll either
have to use another sqlite db or reverse this decision.

Schema version bumped (a lot!)

I meddled with unzipping to make it generic as we now do it for two
different files.

We can now add search on the new fields like rarity and artist and fallback
images with multiverseid can now be made to work. That's upcoming.

### DIFF
--- a/database.py
+++ b/database.py
@@ -1,3 +1,4 @@
+import os
 import sqlite3
 
 import pkg_resources
@@ -7,7 +8,7 @@ import configuration
 
 class Database:
     # Bump this if you modify the schema.
-    schema_version = 2
+    schema_version = 16
 
     @staticmethod
     def escape(s):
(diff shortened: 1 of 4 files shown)


database.py
```diff
@@ -119,26 +135,8 @@ def setup(self):
...
-        self.execute('DROP TABLE IF EXISTS card_subtype')
-        self.execute('DROP TABLE IF EXISTS card_supertype')
-        self.execute('DROP TABLE IF EXISTS card_type')
-        self.execute('DROP TABLE IF EXISTS color')
-        self.execute('DROP TABLE IF EXISTS rarity')
-        self.execute('DROP TABLE IF EXISTS version')
-        self.execute('DROP TABLE IF EXISTS db_version')
+    # Drop the database so we can recreate it.
+    def delete(self):
+        os.remove(configuration.get('database'))
```
silasary, line 141: This won't work on Windows.  You've still got the file handle open from earlier.

oracle.py
```diff
@@ -32,11 +33,12 @@ def update_legality(self, legal_cards):
 
     def update_database(self, new_version):
         self.database.execute('DELETE FROM version')
-        self.database.execute('DELETE FROM card')
```
silasary, line 35: Why was this removed?

bakert: It doesn't do anything? Do we ever update the card table without first dropping everything?

silasary: If mtgjson has an update, but our database doesn't, then we go straight into this without dropping the tables.

bakert: We'd need to drop all printing, subtype, etc. info for that to work, wouldn't we? Or is sqlite sophisticated about cascading foreign keys? Seems unlikely?

silasary: We could use UNIQUE keys, and ON CONFLICT REPLACE?

PR #91, opened 2016-10-02
Added a showall command

binarySwordsman:
showall is a private command, meant to only be used when PMing the bot. It shows all results, not just 10.


silasary:
I'm not overly happy with this, but it'll do for now.

bakert:
I take it this was an emergency hack to work around magidex not being updated? :)

Are you working on a more beautiful version @binarySwordsman?

The time has probably come to make a Command class that can be inherited from.

PR #94, opened 2016-10-02
Refactor commands

silasary:


bakert (commented):
Looooove this. Glorious and overdue :)

You made Commands a class because you wanted to be able to loop over its methods? I'm a full convert to module-level functions but we're not referring to this anywhere else so if the looping is easier this way I guess it doesn't matter.

Nitpick: spaces,after,commas,in,args,lists.

silasary:
It's mostly so I can have the handle_command in the same file without any possibility of a recursive call.

PR #115, opened 2016-10-04
Functional tests

silasary:


silasary:
It turns out that two of our commands weren't working, and we had no idea.

This adds basic smoke testing, which:
- Mocks up fake Client and Channel instances
- Calls each command (Except !restartbot) with `!cmd args`
- And checks that each one outputs something to the channel without crashing.

PR #116, opened 2016-10-04
Card alias auto-update

Pseudodude:
- Card alias table now auto-updates on startup if record count differs
  from lines in card_aliases.txt
- Added a few requested card aliases


bakert:
Is this why you were asking about running the linter @Pseudodude? Looks like you have a couple of easy-to-fix lint issues.

PR #126, opened 2016-10-08
!p1p1

Volvary:
New request, hopefully fixing the Jenkins error

### DIFF
--- a/bot.py
+++ b/bot.py
@@ -35,12 +35,12 @@ class Bot:
     async def respond_to_command(self, message):
         await command.handle_command(message, self)
 
-    async def post_cards(self, cards, channel, additional_text='', verbose=False):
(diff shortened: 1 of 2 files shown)


bot.py
```diff
@@ -35,12 +35,12 @@ def init(self):
     async def respond_to_command(self, message):
         await command.handle_command(message, self)
 
-    async def post_cards(self, cards, channel, additional_text='', verbose=False):
+    async def post_cards(self, cards, channel, additional_text='', verbose=False, legality=True, multirow=False):
```
bakert, line 38: You're probably not going to want to hear this but the real answer to the linter saying "there are too many args" is to use ImageMagick to composite a single image for the draft command rather than have it send multiple images. This will be quicker, make the code (here) simpler and mean that the draft pack will never be broken apart by other people talking.

bakert:
I'm guessing this isn't happening :)

PR #217, opened 2016-10-19
Use card_legality for PD legality the same way we do Legacy, Vintage, etc.

bakert:
This removes the pd_legal column from the card table.

I also made the dubious decision to start storing the legal cards list in mixed
case during this. I will revisit soon with a plan for rules about when and how
we canonicalize.


silasary:
Great :)

PR #264, opened 2016-10-24
No more downloading prices.db.

### LINKED ISSUE #246: updateprices is so slow I've excluded it from testing
We should either make a smaller db with just the cache table in to download, or something more sophisticated.

silasary:
Now we can request the individual card's price (in JSON form).

Fixes #246

### DIFF
--- a/bot.py
+++ b/bot.py
@@ -1,12 +1,8 @@
-import os
-import time
-
 import discord
 
 import command
 import configuration
 import emoji
-import fetcher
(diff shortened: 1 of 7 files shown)


silasary:
Requires that @bakert runs srv_prices.py on his server.

silasary:
(If that's not running, it'll fall back to the ugly existing method, but we really don't want that)

PR #267, opened 2016-10-25
Major refactor

silasary:


silasary:
@bakert Can I have a Code Review?

command.py
```diff
@@ -9,15 +9,11 @@
...
-import database
 import emoji
-import fetcher
-import oracle
 import price
-import rotation
 
-from card import Card
+from magic import configuration, database, oracle, fetcher, rotation
+from magic.card import Card
```
bakert, line 16: I have sympathy for the position that `from X import Y` is a bad practice. If this were magic.card we could have local variables called card without the linter complaining.

Then again magic.card.blah.whatever gets old very quickly. So I could see us going either way. This project is small enough that we don't have huge issues with this kind of import and we have a bunch of top level modules even after this refactor so I'm not sure it matters.

magic/database.py
```diff
@@ -135,7 +135,8 @@ def setup(self):
         self.execute("""CREATE TABLE IF NOT EXISTS fetcher (
             id INTEGER PRIMARY KEY,
             resource TEXT UNIQUE ON CONFLICT REPLACE NOT NULL,
-            last_modified TEXT
+            last_modified TEXT,
+            content TEXT
```
bakert, line 139: Nice.

magic/fetcher.py
```diff
@@ -0,0 +1,47 @@
+import csv
+import json
+
+import pkg_resources
+
+from magic import configuration
+from magic.fetcher_internal import (FetchException, fetch, fetch_json, store,
```
bakert, line 7: Here's where the from gets heavy. But I think it's ok. We're a long way from the list of 200 imports at the top of most of the java files I work on in the day job.

magic/fetcher_internal.py
```diff
@@ -1,39 +1,16 @@
...
-
-def all_cards():
-    s = unzip('https://mtgjson.com/json/AllCards-x.json.zip', 'AllCards-x.json')
-    return json.loads(s)
+from magic import database
 
-def all_sets():
-    s = unzip('https://mtgjson.com/json/AllSets.json.zip', 'AllSets.json')
-    return json.loads(s)
+SESSION = requests.Session()
```
bakert, line 13: These module-level variables that are instantiated when the module is imported freak me out a little bit. For a while we were double fetching whatsinstandard on startup because of the stuff I added in rotation. I don't know python standards well enough to know if I should get over my squeamishness or respond to it. It's sort of like a member variable or a static member variable but it's getting instantiated at import time which I find weird. I can probably cope.

Why is it good to have a Session? Some kind of caching?

bakert (commented):
Good stuff. shipit!

magic/fetcher_internal.py
silasary: Session stores cookies, which I needed for TappedOut browsing.

PR #272, opened 2016-10-25
Added "Hot Potato" as an alias

silasary:
We really need a better way of adding aliases.


bakert:
I'd like !setalias but then the bot is priced into persisting data which we've been resisting mightily!

PR #278, opened 2016-10-26
Decksite

silasary:


bakert:
This is really a very big pile. I've pushed on with decksite and fixed enough that this is mergeable into master.

It doesn't really add anything to master as decksite is not fully functional but I think we should go ahead and merge it anyway and "fix forward".

It doesn't break the bot and that's the main thing we care about.

I'm going to leave it to you @silasary to push the big green button though.

bakert:
Also I went moderately deep on the database. I'm trying to make what we're storing complete enough that we can pull in Gatherling and MTGG decklists without having to refactor out internals just write another scraper/massager to get the data into this schema. Let me know if you think it's off. I had big struggles with what exactly uniquely identifies a deck. I went with URL + a field I'm calling 'identifier'. At times I had source_id, person_id and other things in there. I'm looking for what we can query on when we have a deck and it'll find the existing deck if it is the same deck and it won't if it's not. I'm not even sure what constitutes the _same_ deck. Is it the same deck just because it's the same 75? What about same 75 + same pilot? What about same 75 + same pilot + same tournament but reported in a different place? That last one in particular is a problem for how I resolved it. I did start with "deck" (the 75 cards in the abstract) and "deck instance" (someone publishing or playing with those 75) but that seemed like madness. Let me know what you think anyway.

bakert:
We do at least get this far with this branch:

screen shot 2016-10-26 at 11 13 35 pm

PR #405, opened 2016-10-30
More Cleaning of command.py

silasary:


Makefile
```diff
@@ -26,6 +26,13 @@ lint:
 	@find . -name "*.py" | xargs pylint -f parseable
 	@echo
 
+shortlint:
+	@echo
+	@echo "******************************** Lint *****************************************"
+	@echo
+	@find . -name "*.py" | xargs pylint -f parseable -E
```
bakert, line 33: This is great. Will Jenkins break if we make this the only lint command?

silasary: Jenkins doesn't use `make`.  It has its own independant syntax and build plans.

We might want to change that at some stage, but for now we're fine.

PR #414, opened 2016-10-30
The scrapers seem to get broken a lot.  Let's add tests for them.

silasary:
They currently fail -- Fix them!


silasary:
Tappedout was failing because of Nezumi Graverobber, so I pushed the deck responsible to the second page ;)

PR #440, opened 2016-11-01
Use ['inventory'] if we have access to it.

silasary:


silasary:
Tapped out supports maybe and aquire boards.
We're ignoring those.

bakert:
Nice. Thanks!

Issue #458, opened 2016-11-02
Add Season X legality lists to database.


silasary:

bakert:
We did this.

PR #463, opened 2016-11-02
Add ability to fetch older season's legality lists.

silasary:
This is a partial implementation of #458.

Currently, it can grab the older lists, and add them to the formats table.

This code is not called anywhere, and the deck checking code can't check formats other than `"Penny Dreadful"`.

Both of these things should be done at some point.


silasary:
Almost ready with this one.  But I'm heading to bed now.
Need to display the data in a nicer way, and do more in the way of optimizing the initialization of `legality.LEGAL_CARDS`

legal in

We need to do something about card counting, because right now cards that are over the limit (Singleton formats or vintage restricted cards) don't disqualify a format.

I'll think more about that tomorrow.

bakert:
I had a look at this although my brain is not really working. It seems moderately complicated.

I wrote this:

```
+# BAKERT I don't know the deck construction rules of Prismatic.
+def legal_formats(d):
+    if len(d.maindeck) < 60:
+        print('a')
+        return set()
+    if len(d.sideboard) > 15:
+        print('b')
+        return set()
+    card_count = {}
+    highest = 0
+    for c in d.all_cards():
+        if not c.type.startswith('Basic Land'):
+            card_count[c.name] = card_count.get(c.name, 0) + 1
+    if card_count.values() and max(card_count.values()) > 4:
+        print('c')
+        return set()
+    formats = {r['id']: r['name'] for r in db().execute('SELECT name, id FROM format')}
+    if card_count.values() and max(card_count.values()) > 1 or len(d.maindeck) < 99:
+        formats = {format_id: name for format_id, name in formats.items() if name != 'Commander' }
+    legal_cards_by_format = {}
+    rs = db().execute('SELECT card_id, format_id FROM card_legality')
+    for format_id in set(row['format_id'] for row in rs):
+        legal_cards_by_format[format_id] = []
+    for r in rs:
+        legal_cards_by_format[r['format_id']].append(r['card_id'])
+    for c in d.all_cards():
+        for format_id in formats.keys():
+            # BAKERT why is this cast necessary?
+            if int(c.id) not in legal_cards_by_format[format_id]:
+                print('removing {format_id} because {name} is not in it'.format(format_id=format_id, name=c.name))
+                formats.pop(format_id, None)
+                break
+    print('d')
+    return formats.values()
```

Which passes these tests:

```
from munch import Munch

from decksite.data import deck

from magic import legality, oracle

# BAKERT we don't cope with 'Restricted' legality.
def test_legal_formats():
    d = deck.Deck({'id': 0})
    d.maindeck = oracle.load_cards(['Swamp']) * 59
    d.sideboard = []
    assert len(legality.legal_formats(d)) == 0
    d.maindeck = oracle.load_cards(['Swamp']) * 60
    formats = legality.legal_formats(d)
    assert 'Penny Dreadful' in formats
    assert 'Legacy' in formats
    d.maindeck = oracle.load_cards(['Swamp']) * 55 + oracle.load_cards(['Think Twice']) * 5
    assert len(legality.legal_formats(d)) == 0
    d.maindeck = oracle.load_cards(['Swamp']) * 56 + oracle.load_cards(['Think Twice']) * 4
    formats = legality.legal_formats(d)
    assert 'Legacy' in formats
    assert 'Modern' in formats
    d.sideboard = oracle.load_cards(['Swamp']) * 15 + oracle.load_cards(['Think Twice'])
    assert len(legality.legal_formats(d)) == 0
    d.maindeck = oracle.load_cards(['Swamp']) * 56 + oracle.load_cards(['Fork']) * 4
    d.sideboard = oracle.load_cards(['Swamp']) * 15
    formats = legality.legal_formats(d)
    assert 'Legacy' in formats
    assert 'Modern' not in formats
```

But it's still slow and confusing and now I've started hardcoding deck construction rules. Which may be unavailable.

bakert:
Full diff here with a couple of changes elsewhere necessary to make it work: https://gist.github.com/bakert/08249c21dc39a7386beeea406f644d65

silasary:
> Each Prismatic deck must contain at least 250 cards, including at least 20 cards of each color. If a card is multicolored, you choose which of its colors to count it as for this purpose. For example, a red-green card can count as one of your 20 red cards or one of your 20 green cards, but not both.

Prismatic goes on our "Easier to ignore" list.

silasary:
screenshot 2016-11-03 17 31 38

silasary:
Now all I need to do is the add the `/decks/` **Legality** column, and make Jenkins happy again.

bakert:
Merging this in its imperfect state because it at least broadly works. We can improve from here.

PR #480, opened 2016-11-02
DO NOT MERGE INCOMPLETE First stab at adding leagues/league sign up.

bakert:


bakert:
I merged this even though it's not complete because it works and it's not exposed in the menu or anything.

PR #488, opened 2016-11-03
Trying to make legality a little cleaner/more efficient.

bakert:


bakert:
Ok, we have tests and the ability to check size of sideboards and maindecks, and 4-ofs, plus we load card legality when we load the cards so we don't need separate loading of what is legal. I'm not sure why I never thought of grouping the inner query in base_query on face.id before - it will make other things possible too. We can probably go further on that and remove the EMN list but I didn't look into that yet.

PR #567, opened 2016-11-07
Couldn't upload to league when the deck contained a card in both the mainboard and sideboard.

silasary:


silasary:
Fixes bug reported by Mathgeek on Discord.

PR #578, opened 2016-11-07
Improved the feel of repeated elements.

silasary:


bakert (approved):
You can delete rather than commeting. git will remember :)

PR #579, opened 2016-11-07
Make the results and announcement links work on signup and report.

### LINKED ISSUE #568: Links on league reporting page are self-referential
The links labeled "League Announcement" and "League Results" both point back to the league reporting page

bakert:
Fixes #568.

Standardize on 'competition_url' for our page and 'url' for the source.

### DIFF
--- a/decksite/data/competition.py
+++ b/decksite/data/competition.py
@@ -31,7 +31,7 @@ def load_competition(competition_id):
 
 def load_competitions(where_clause='1 = 1'):
     sql = """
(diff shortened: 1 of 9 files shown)


bakert (commented):
Because this uses load_competition it now completely unnecessarily loads all the decks in the competition. This kind of thing may or may not make the site slow but I'd rather start with developer friendliness and increase perf than worry about it in advance. We should measure speed at some point, though, and improve the slow parts.

PR #587, opened 2016-11-08
I'm updating Jenkins, and they changed the syntax slightly.

silasary:
https://github.com/jenkinsci/pipeline-model-definition-plugin/wiki/Version-History-and-Changes#version-05


silasary:
Well, at some point the normal tests will stop working, and this will start working.  When it does, merge it.

PR #595, opened 2016-11-11
Scraper upgrades

silasary:


silasary:
Ah, that's what happens when I turn that on.  Neat.

silasary:
So `/add/` is now a text box that asks for a URL.
It's ugly, doesn't handle errors well, and isn't accessible from anywhere.
But it's good enough for the discord bot.

silasary:
We should probably verify that decks are remotely PD legal before adding them to the DB.
People occasionally post standard or Modern decks in chat, and we don't want to accidentally add them here.

PR #603, opened 2016-11-14
Correctly clean up before a mtgjson update.

silasary:


silasary:
Fixes the bug that broke PROD the other day.

PR #609, opened 2016-11-16
Hash decklists

### LINKED ISSUE #293: add_deck should possibly update the deck and it's updated_date instead of just returning deck_id when sent an existing deck

silasary:
We now:
- hash decklists
- can update the contents of a tappedout decklist.  (It turns out that we were previously just dropping them. Fixes #293)


silasary:
Fixes #293

silasary:
Multiple runs

silasary:
It's not perfect - stars and top8 are showing the main-deck info, not the twin's info - But it's getting there.

PR #639, opened 2017-01-02
Migrate decksite database to mysql.

### LINKED ISSUE #626: Switch to MySQL.
Sqlite has hit the boundaries of what we can support.

We either need to rethink the way we execute scrapers, or we need to move to a database that supports multithreaded access.

Moving to MySQL feels like the easier of the two options.

silasary:
Leaving cards.sqlite as is for now, because it's a mess and doesn't suffer race conditions anywhere near as often.
NB: Tests need to be fixed before this is mergable.
Closes #626


bakert:
Ooh, exciting.

silasary:
Any idea why pylint is complaining about similar lines in `__init__.py`?
http://ci.katelyngigante.com/job/Penny%20Dreadful/job/Penny-Dreadful-Tools/job/mysql/21/warnings48Result/

PR #676, opened 2017-04-29
Update Decksite legality indicators

silasary:
This still needs to be automated.


silasary:
Travis is failing because there's a tournament that is set up very wrong.  Penny Paradise 2.01 starts on `30 November -0001`.

PR #692, opened 2017-05-06
Accomodate new columns in bugged cards spreadsheet.	

bakert:


bakert:
This is irrelevant now we use modo-bugs.

PR #727, opened 2017-05-14
Actually implement some API stuff.

silasary:
Nowhere near complete, and probably exposes more data than is considered useful.

But it's a start, and provides me with the basis on which to add League Support to PDBot.

### DIFF
--- a/decksite/__init__.py
+++ b/decksite/__init__.py
@@ -1,0 +1,2 @@
+from flask import Flask
+APP = Flask(__name__)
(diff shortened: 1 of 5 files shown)


bakert (commented):
Well, this is awesome :)

I don't have too much feedback I'd just say make sure you develop it in conjunction with what the bot actually needs to make sure the API "works". But seems like you're doing that anyway. And it's less important here because the API only has one client for now, but important before it gets locked in.

Sweet stuff.

bakert:
Is the PDBot code in github/PDT too?

silasary:
No, PDBot is unfortunately closed source.

The way I'm interfacing with the MTGO client is not something I can share, and I don't believe that the rest of the code would be useful without that.

PR #731, opened 2017-05-16
Correctly insert the current time when creating matches.  

### LINKED ISSUE #730: Match dates are set incorrectly.
Reported on Discord by silasary#2393

silasary:
Fixes #730

### DIFF
--- a/decksite/league.py
+++ b/decksite/league.py
@@ -1,4 +1,5 @@
 import json
+import time
 import datetime
 import calendar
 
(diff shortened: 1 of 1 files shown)


decksite/league.py
```diff
@@ -132,7 +133,7 @@ def active_league():
     return guarantee.exactly_one(leagues)
 
 def insert_match(params):
-    match_id = db().insert("INSERT INTO `match` (`date`) VALUES (DATE_FORMAT(now(), '%%s'))")
```
bakert, line 135: Could also use UNIX_TIMESTAMP() to keep it (my)SQL. Does this work the same regardless of timezone? I assume so but if not we could also use our home grown dtutils.dt2ts(dtutils.now()).

PR #735, opened 2017-05-18
Querying split cards and other cleanup.

bakert:


magic/multiverse.py
```diff
@@ -259,274 +258,3 @@ def card_name(c):
...
-            {"format": "Legacy", "legality": "Legal"},
-            {"format": "Vintage", "legality": "Legal"},
-            {"format": "Commander", "legality": "Legal"}],
-        "name": "Budoka Gardener",
-        "subTypes": ["Human", "Monk"],
-        "printings": ["CHK"],
-        "colors": ["Green"],
-        "rarity": "Rare"
-    }
-    cards["Gleemox"] = {
```
silasary, line 519: One of these is not like the others.
We need to hardcode Gleemox because it isn't (and won't be) added to mtgjson.

silasary (changes requested):

magic/multiverse.py
bakert: Ooh, good spot. Thanks!

PR #813, opened 2017-06-04
Charts of CMC on deck page.

bakert:


silasary:
Add the offending modules to `[TYPECHECK] ignored-modules=` in the pylintrc file

PR #895, opened 2017-06-22
Welcome msg

silasary:


magic/database.py
```diff
@@ -104,14 +104,14 @@ def column_def(name, prop):
         primary_key = 'PRIMARY KEY' if prop['primary_key'] else ''
         default = 'DEFAULT {default}'.format(default=prop['default']) if prop['default'] is not None else ''
         unique = 'UNIQUE' if prop['unique'] else ''
+        if prop['type'].startswith('VARCHAR'):
+            prop['type'] = 'TEXT'
```
bakert, line 108: What's this for?

silasary: Long story short, Sqlite doesn't support VARCHAR, and Mysql claims that the max row size of `face` is absurd if they're all TEXT.

This allows me to manually shorten columns (like `name`) that will never be too large for a VARCHAR(200).

All part of the ongoing "Let's get this thing working on mysql"

bakert: Ah, I see. Sounds good.

PR #918, opened 2017-06-26
Rotation error

silasary:
Include some sanity checks on /rotation/.


silasary:
Diagnoses, but does not fix #919

PR #978, opened 2017-07-06
Redact rotation stats on undecided cards.

silasary:


silasary:
Redacted numbers actually look nice.

&block;  is a great character

silasary:
Note that we do redact based on string length, so people can see the difference between 1 and 2 digit numbers.

PR #981, opened 2017-07-07
Fixing up detail pages.

bakert:


bakert:
Don't merge this yet- still thinking about it.

PR #1002, opened 2017-07-12
Improve the handling and display of archetypes and archetype detail page.

bakert:
(1) Simplify tree code by using the anytree library instead of homespun code.
(2) Use the anytree NodeMixin to make Archetypes and tree entries one and the
    same thing (so we don't have to `update` each tree entry to get the
    archetype properties).
(3) Make Archetype a class instead of using Container so we can use NodeMixin.
(4) Load parent information when querying for archetypes in order to use
    anytree's features.
(5) ORDER BY number of decks by default so output from anytree shows more
    popular archetypes first among archetypes at the same level. ORDER BY name
    for the dropdown on edit archtypes page still.
(6) Remove load_tree because anytree does that for us by traversing using
    parent.
(7) Add a load_archetypes_without_deck_with_root helper function to find any
    given archetypes slice of the whole tree without gumming up main.py.
(8) Use load_archetypes_without_deck_with_root results to display the correct
    slice of the archetype tree on the archetype detail page.
(9) Highlight the currently viewed archetype in the tree slice on the
    archetype detail page with a light gray background.
(10) Replace homespun preorder code with anytree's.
(11) Remove sorting. Combination of preorder and database ORDER BY now gets us
     that for free.
(12) Add dependency on anytree to requirements.txt

### DIFF
--- a/decksite/data/archetype.py
+++ b/decksite/data/archetype.py
@@ -44,7 +44,7 @@ def load_archetypes(where_clause='1 = 1', merge=False):
     archetypes = list(archetypes.values())
     return archetypes
 
-def load_archetypes_without_decks(where='1 = 1', order_by='`season.num_decks` DESC, `all.num_decks` DESC, `season.wins` DESC, `all.wins` DESC'):
+def load_archetypes_deckless(where='1 = 1', order_by='`season.num_decks` DESC, `all.num_decks` DESC, `season.wins` DESC, `all.wins` DESC'):
     sql = """
         SELECT
             a.id,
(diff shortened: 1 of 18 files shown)


silasary:
Well, codacy doesn't seem to like this.

bakert:
Codacy doesn't like the new imports, which is weird. We saw that recently when I added a dependency, too. Even weirder it doesn't like `import pytest` which is working fine in scraper_test.py and run.py. Hm.

bakert:
It was the imports in 1183496 that were the issue last time. Libraries for drawing the charts.

bakert:
From your note on the previous pull request I think I need to add inflect and anytree to `ignored-modules=apsw,munch,matplotlib,seaborn`. Buuuut that doesn't explain the pylint issue. Perhaps I don't need that because it's loaded in run.py anyway.

PR #1211, opened 2017-08-02
Squash people and fixes.

bakert:


magic/fetcher.py
```diff
@@ -100,3 +100,7 @@ def decksite_url(path='/'):
     if port != 80:
         hostname = '{hostname}:{port}'.format(hostname=hostname, port=port)
     return parse.urlunparse((configuration.get('decksite_protocol'), hostname, path, None, None, None))
+
+def cardhoarder_url(d):
+    deck_s = '||'.join([str(entry['n']) + ' ' + entry['card'].name.replace(' // ', '/').replace('"', '') for entry in d.maindeck + d.sideboard])
+    return 'https://www.cardhoarder.com/decks/upload?deck={deck}'.format(deck=internal.escape(deck_s))
```
silasary, line 106: We're missing an affiliateID here.

bakert: We don't have one. It came up briefly in my conversation but he said the commission would be insignificant and they are doing us a favor so I didn't pursue it.

silasary: Sure

PR #1281, opened 2017-08-22
Argh, I broke things on production. 

silasary:
Apparently View is a complex and messy thing that shouldn't be imported lightly.

### DIFF
--- a/decksite/views/tournaments.py
+++ b/decksite/views/tournaments.py
@@ -1,48 +1,35 @@
-from dateutil import rrule
-
-from shared import dtutil
+from magic import tournaments
 
 from decksite.view import View
 
 # pylint: disable=no-self-use
 class Tournaments(View):
     def __init__(self):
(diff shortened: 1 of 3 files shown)


decksite/views/tournaments.py
```diff
@@ -1,48 +1,35 @@
...
-        elif pdt_time < pds_time:
-            self.next_tournament_name = 'Penny Dreadful Thursday'
-            next_time = pdt_time
-        else:
-            self.next_tournament_name = 'Penny Dreadful Sunday'
-            next_time = pds_time
-        self.next_tournament_time_precise = dtutil.dt2ts(next_time) - now_ts
-        self.next_tournament_time = dtutil.display_time(self.next_tournament_time_precise, granularity=1)
+        
+        info = = tournaments.next_tournament_info()
```
bakert, line 9: This double equals looks like a bug?

silasary: Yup.  I addressed it in #1282

Issue #1409, opened 2017-10-15
Tighten up /retire/ so there can only be one MTGO username per discord_id and so we prompt you with your own deck and so there is a link somewhere and so there is a link to logout (but don't break privacy with server-based full page caching)


vorpal-buildbot:
Reported on Discord by bakert#2193

jgabrielygalan:
Related to #3778, not sure if that one is enough to cover everything is expected from this one.

bakert:
Oh, I see I covered my concerns about 3778 here :)

The one thing missing is that once you retire we should send you to sign up not to a 404. Not sure if that is covered in another ticket.

jgabrielygalan:
Ah, yes, but only if it's actually deleted when you had 0 matches. I'll open one.

PR #1410, opened 2017-10-15
Rudimentary retire-via-discord auth.

bakert:
This is not totally secure but any action you take here can be reversed and
we can work out who you are so it's secure enough for now.

Obvious potential improvements are obvious.


bakert:
See #1409 for the stuff I'd do first to improve this.

PR #1599, opened 2017-11-11
Make the About page more interesting.

### LINKED ISSUE #1489: It would be nice to programmatically generate the names of cards in /about/ so we don't display illegal cards there as seasons go on. but i'm not sure what algorithm could be used to generate it.

silasary:


silasary:
Feel free to add more to that list if you think of fun exciting cards.

PR #1613, opened 2017-11-12
More design work.

bakert:
In an attempt to make our pages with body text more legible:

    1. Split 'about penny dreadful' and 'about pdm' into two pages.
    2. Add dropdown on ABOUT to accomodate such.
    3. Change from Golden Ratio (1.618) to Perfect Fourth (1.333) so that
       our line-height can match the range suggested by Practical Typography    4. Make all "thin" and "quite thin" lines instead into a uniform 1px
       (in rem).
    5. Add more notes on what we think it is we are doing design-wise to the
       CSS.
    6. Change body text AND headings to use new font and drop Lato.
    7. Inline said font in CSS.
    8. Add some more notes on how we calculate various values. This stuff
       is crying out to be in SASS. I'll do that at some point.


bakert:
Some of this is not optimal. /tournaments/ for example is now uglier than before. But we'll get there. Releasing this just because I don't want tons of stuff queueing.

PR #1946, opened 2017-11-19
Caching of cards.

bakert:


bakert:
This test passes on local despite:

    sqlite> SELECT layout FROM _cache_card;
    Error: no such column: layout

which is a bit weird.

bakert:
The problem is we somehow end up with these columns:

    sqlite> .schema _cache_card
    CREATE TABLE _cache_card(
      "`u`.`layout`" TEXT,
      "`u`.`id`" INT,
    ...

PR #2050, opened 2017-11-22
decksite: smarter algorithm to guess sideboard from maindeck, fixes #1554

### LINKED ISSUE #1554: Sometimes league decks have their sideboard put in the maindeck
https://pennydreadfulmagic.com/decks/5229/

We need defensive code and maybe logging for when this gets past the defensive code.

Is there anything in common between the effected decks? 74 cards?

jgabrielygalan:
The algorithm tries to do the following:
- If the list contains empty lines or the word "sideboard" at the beginning of a line, the last block is assigned as sideboard and the rest as maindeck.
- In other case (1 continuous list of cards): gather 60 cards for maindeck from the beginning, then gather 15 cards for sideboard starting from the end, then the rest to maindeck

It supports less than 15 sideboards cards.
It implies that a deck with more than 60 cards needs to have a 15 card sideboard, or some maindeck cards could go to the sideboard.


decksite/scrapers/decklist.py
```diff
@@ -8,37 +8,53 @@
...
-            if d['maindeck'][name] == 0:
-                del d['maindeck'][name]
-    return d
+    s = s.lstrip().rstrip()
+    maindeck = {}
+    sideboard = {}
+    chunks = re.split('\r?\n\r?\n|^\s*sideboard.*?\n', s, flags=re.IGNORECASE|re.MULTILINE)
+    if len(chunks) > 1:
+        for chunk in chunks[:-1]:
+            parse_chunk(chunk,maindeck)
```
bakert: nit: space after comma

decksite/scrapers/decklist.py
```diff
@@ -8,37 +8,53 @@
...
+        parse_chunk(chunks[-1],sideboard)
```
bakert: And everywhere else I guess :)

decksite/scrapers/decklist_test.py
```diff
@@ -165,3 +166,97 @@ def test_parse5():
...
+        4 Scrabbling Claws
+    """
+    s = textwrap.dedent(s)
+    d = decklist.parse(s)
+    print(d)
+    assert sum(d['maindeck'].values()) == 61
+    assert len(d['maindeck']) == 13
+    assert len(d['sideboard']) == 7
+
+# Test a 63 card deck + 12 sideboard
```
bakert, line 201: I want to smack whoever does this :)

decksite/scrapers/decklist_test.py
```diff
@@ -165,3 +166,97 @@ def test_parse5():
...
+        1 Lone Revenant
+        1 Careful Consideration
+        1 Opportunity
+        4 Annul
+        4 Invasive Surgery
+        3 Sentinel Totem
+    """
+    s = textwrap.dedent(s)
+    d = decklist.parse(s)
+    print(d)
```
bakert, line 258: I'm not sure why we print these things but I guess the other tests do it too!

bakert (approved):
Awesome. Thanks!

decksite/scrapers/decklist_test.py
jgabrielygalan: It seems I copied it from test_5 without noticing...

decksite/scrapers/decklist_test.py
jgabrielygalan: Haha, well I wanted to test that when adding to the sideboard bottom up, it actually stopped when adding the N cards from a row would exceed 15

PR #2072, opened 2017-11-23
Disable tooltips on touch devices.

bakert:


bakert:
This probably isn't safe to merge. There may be a race condition in the javascript loading.

Issue #1617, opened 2017-11-12
Don't store a sixth result in the league whether it came from pdbot getting confused, a stale report form or anything else. Just error, nicely.


vorpal-buildbot:
Reported on Discord by bakert#2193

bakert:
@jgabrielygalan did this.

PR #2101, opened 2017-11-23
Reject sixth report

jgabrielygalan:
As explained in discord: in the report method I'm checking that neither you nor your opponent have already 5 matches reported. It wraps the whole check + update in a distributed lock if the DB supports it. If we want to quickly deactivate the lock the easiest way would be to change database_mysql.py to return False in the supports_lock method.


jgabrielygalan:
Fixes #1617

silasary (approved):
Looks good to me

PR #2102, opened 2017-11-23
Version compare compat

jgabrielygalan:
I was having problems with the database update checker. The culprit was that different versions of setuptools had different return types for the parse_version method. This commit cleans up the usage of str of the returned object in favor of storing the version coming from mtgjson, and use pkg_resources.parse_version only for the comparison.

### DIFF
--- a/magic/database.py
+++ b/magic/database.py
@@ -21,7 +21,7 @@ def init():
         setup()
 
 def version() -> str:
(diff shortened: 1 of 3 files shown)


jgabrielygalan:
Don't merge this yet. The codacy report showed me a mistake I need to fix

Issue #1606, opened 2017-11-11
Bot should know that 1 hour after a tournament starts that there is a tournament going on and hint as much to !tournament command


vorpal-buildbot:
Reported on Discord by bakert#2193

jgabrielygalan:
Do you mean on top of reporting the next one, mentioning that a recent one has just started?

silasary:
"Fixes #XXX" must be in the commit message or the body of a PR to work.

PR #2854, opened 2017-11-28
discordbot: report previous tournament date

jgabrielygalan:
In the !tournament command, the bot will now tell when the previous tournament took place or number of hours ago it started if it's less than 5.


jgabrielygalan:
Fixes: #1606

Issue #1633, opened 2017-11-13
Bug symbol should show details of bug on hover


vorpal-buildbot:
Reported on Discord by bakert#2193

bakert:
As this is added in :after by CSS we would either have to change how it works, dynamically generate the CSS, or attach it with javascript. All of those are fairly reasonable options. As it will be a tooltip anyway (which uses js) I think the best solution is probably to do it in js.

PR #2876, opened 2017-11-28
decksite: tooltip with the bug description for bugged cards

jgabrielygalan:
Fixes: #1633.

### DIFF
--- a/decksite/static/css/pd.css
+++ b/decksite/static/css/pd.css
@@ -411,6 +411,10 @@ ul.decklist {
     font-size: 0.75rem;
 }
 
+ul.decklist li a {
+    float:left;
+}
+
 /*
 
 Tables
@@ -859,9 +863,8 @@ p.bugged {
 }
 
 
-.bugged::after {
-    content: "🐞 ";
-    margin-left: 0.2em;
+.bugicon {
+    float:left;
 }
 
 .illegal::after {
--- a/decksite/static/css/pd.css
+++ b/decksite/static/css/pd.css
@@ -862,9 +862,13 @@ p.bugged {
     width: 21.034rem;
 }
 
+.decklist li {
+    clear: both;
+}
 
 .bugicon {
     float:left;
+    margin-left: 0.2em;
 }
 
 .illegal::after {
--- a/decksite/static/css/pd.css
+++ b/decksite/static/css/pd.css
@@ -412,10 +412,6 @@ ul.decklist {
     font-size: 0.75rem;
 }
 
-ul.decklist li a {
-    float:left;
-}
-
 /*
 
 Tables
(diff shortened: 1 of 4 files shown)


decksite/static/js/pd.js
```diff
@@ -162,6 +163,14 @@ PD.showLocalTimes = function () {
         $(this).html(t.tz(moment.tz.guess()).format("dddd h:mma z")).parent(".local").show();
     });
 }
+PD.initBugIcons = function() {
+    $(".card.bugged").each(function() {
+        var element = $("<div class=\"bugicon\"> 🐞 </div>")
+        element.insertAfter($(this));
+        Tipped.create(element, $(this).attr("data-bug-desc"), {maxWidth: "200"});
+    });
+}
```
bakert: Does doing it in js cause the bug icons to "blink" into the page after the page has already rendered? /bugs/ is a good place to check.

decksite/templates/entry.mustache
```diff
@@ -1 +1 @@
-<li><a href="{{card.url}}" class="{{^card.pd_legal}}illegal{{/card.pd_legal}} {{#card.bug_desc}}bugged{{/card.bug_desc}} card">{{n}} {{name}}</a></li>
+<li><a href="{{card.url}}" class="{{^card.pd_legal}}illegal{{/card.pd_legal}} {{#card.bug_desc}}bugged{{/card.bug_desc}} card" {{#card.bug_desc}}data-bug-desc="{{card.bug_class}}: {{card.bug_desc}}"{{/card.bug_desc}}>{{n}} {{name}}</a></li>
```
bakert: We could write out the bug element with a title attribute here and the existing tooltip-adding code would add the tooltip and get rid of the need for initBugIcons?

decksite/static/css/pd.css
```diff
@@ -859,9 +863,8 @@ p.bugged {
 }
 
 
-.bugged::after {
-    content: "🐞 ";
-    margin-left: 0.2em;
+.bugicon {
+    float:left;
```
bakert: If we don't clear anywhere does this cause cards with shorter names to bunch up more-than-one on a line?

Do we still need the margin-left to stop the bug icon touching the associated card? We want a small gap there.

bakert (commented):
Cool stuff. Some suggestions but we can ship as-is if you're happy with it.

decksite/templates/entry.mustache
jgabrielygalan: I think you are right. I was obssessed with creating everything in js, I'll try to do as you say here.

decksite/static/js/pd.js
jgabrielygalan: With the change of strategy of generating the div with the bug icon in mustache, this doesn't happen, so no problem here.

decksite/static/css/pd.css
jgabrielygalan: There was already a clear:both in the .decklist that I thought would be enough, I'm not sure. I'll add the clear:both to the li too, just in case.

jgabrielygalan: And you were right too with the margin-left :)

bakert:
After my two little changes:

captura de pantalla 2017-11-29 a las 8 11 39 a m

PR #2978, opened 2017-11-30
Improve fuzzy matching behavior.

bakert:


silasary (commented):
Something in this commit caused the searchbox issues.

bakert:
What are the searchbox issues? Do we have something logged?

PR #3363, opened 2017-12-05
Google bot command. Fixes #1555

### LINKED ISSUE #1555: Googler not working from prod box

jgabrielygalan:
Replace googler with google-api-python-client to use the Custom Search Google API. We need to follow these instructions to obtain an API Key and a Custom Search Engine ID: 

https://stackoverflow.com/a/37084643/2773639

and place them in the config.json as cse_api_key and cse_engine_id.

Also check these caveats about the API and the quotas, pricing: 
https://developers.google.com/custom-search/json-api/v1/overview?hl=es
100 requests per day are free. I'm not sure how much is this feature used.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -282,7 +282,7 @@ ignored-modules=anytree,apsw,bs4,cachecontrol,dateutil,inflect,matplotlib,markdo
 # List of class names for which member attributes should not be checked (useful
 # for classes with dynamically set attributes). This supports the use of
 # qualified names.
-ignored-classes=optparse.Values,thread._local,_thread._local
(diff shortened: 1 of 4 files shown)


pylintrc
```diff
@@ -282,7 +282,7 @@ ignored-modules=anytree,apsw,bs4,cachecontrol,dateutil,inflect,matplotlib,markdo
 # List of class names for which member attributes should not be checked (useful
```
silasary, line 282: Add `googleapiclient` to `ignored-modules`.

silasary (changes requested):

PR #3770, opened 2017-12-08
Decouple the action of populating multiverse from loading oracle.

silasary:


silasary:
Needs a pytest_configure() hook to initialize multiverse.

PR #3857, opened 2017-12-22
Very hopeful attempt to get codacy to behave differently.

bakert:


silasary:
How does this work?

bakert:
The makefile uses a different, temporary pylintrc that it generates from the "real" pylintrc but removes the "bad" line and replaces it with something less severe that will run fine locally and check more stuff. Basically ignored-modules in our pylintrc is now for codacy only.

PR #3879, opened 2017-12-29
decksite: rotation changes page

jgabrielygalan:
New page showing what cards are new to the season and which ones left (were legal last season but not anymore). Linked from the Resources - Deckbuilding section.

### DIFF
--- a/decksite/main.py
+++ b/decksite/main.py
@@ -15,7 +15,7 @@ from decksite.cache import cached
 from decksite.data import archetype as archs, card as cs, competition as comp, deck as ds, person as ps, query
 from decksite.charts import chart
 from decksite.league import ReportForm, RetireForm, SignUpForm
-from decksite.views import About, AboutPdm, AddForm, Archetype, Archetypes, Bugs, Card, Cards, Competition, Competitions, Deck, Decks, EditArchetypes, EditMatches, Home, InternalServerError, LeagueInfo, NotFound, People, Person, Prizes, Report, Resources, Retire, Rotation, RotationChecklist, Season, Seasons, SignUp, TournamentHosting, TournamentLeaderboards, Tournaments, Unauthorized
+from decksite.views import About, AboutPdm, AddForm, Archetype, Archetypes, Bugs, Card, Cards, Competition, Competitions, Deck, Decks, EditArchetypes, EditMatches, Home, InternalServerError, LeagueInfo, NotFound, People, Person, Prizes, Report, Resources, Retire, Rotation, RotationChecklist, Season, Seasons, SignUp, TournamentHosting, TournamentLeaderboards, Tournaments, Unauthorized, RotationChanges
 
 # Decks
 
@@ -251,6 +251,13 @@ def do_claim():
         lg.retire_deck(d)
         return redirect(url_for('signup'))
     return retire(form)
+
+
+@APP.route('/rotationchanges')
+def rotation_changes():
+    view = RotationChanges(*oracle.last_pd_rotation_changes())
+    return view.page()
+
 
 # Admin
 
--- a/decksite/templates/entry.mustache
+++ b/decksite/templates/entry.mustache
@@ -1,5 +1,5 @@
 <li>
-    <a href="{{card.url}}" class="{{^card.pd_legal}}illegal{{/card.pd_legal}} card">{{n}} {{name}}</a>
+    <a href="{{card.url}}" class="{{^card.pd_legal}}illegal{{/card.pd_legal}} card">{{#n}}{{n}} {{/n}}{{name}}</a>
     {{#card.bugs}}
         <a href="{{card.url}}" class="bug" title="{{description}}: {{classification}}">🐞</a>
     {{/card.bugs}}
--- a/decksite/views/rotation_changes.py
+++ b/decksite/views/rotation_changes.py
@@ -1,0 +1,14 @@
+from decksite.view import View
+
+# pylint: disable=no-self-use
+class RotationChanges(View):
+    def __init__(self, cards_in, cards_out):
+        self.sections = []
+        self.cards = cards_in + cards_out
+        entries_in = [{'n': None, 'name': c.name, 'card': c} for c in cards_in]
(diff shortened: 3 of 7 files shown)


decksite/main.py
```diff
@@ -15,7 +15,7 @@
 from decksite.data import archetype as archs, card as cs, competition as comp, deck as ds, person as ps, query
 from decksite.charts import chart
 from decksite.league import ReportForm, RetireForm, SignUpForm
-from decksite.views import About, AboutPdm, AddForm, Archetype, Archetypes, Bugs, Card, Cards, Competition, Competitions, Deck, Decks, EditArchetypes, EditMatches, Home, InternalServerError, LeagueInfo, NotFound, People, Person, Prizes, Report, Resources, Retire, Rotation, RotationChecklist, Season, Seasons, SignUp, TournamentHosting, TournamentLeaderboards, Tournaments, Unauthorized
+from decksite.views import About, AboutPdm, AddForm, Archetype, Archetypes, Bugs, Card, Cards, Competition, Competitions, Deck, Decks, EditArchetypes, EditMatches, Home, InternalServerError, LeagueInfo, NotFound, People, Person, Prizes, Report, Resources, Retire, Rotation, RotationChecklist, Season, Seasons, SignUp, TournamentHosting, TournamentLeaderboards, Tournaments, Unauthorized, RotationChanges
```
bakert, line 18: Nit. These are in alphabetical order.

decksite/main.py
```diff
@@ -252,6 +252,13 @@ def do_claim():
         return redirect(url_for('signup'))
     return retire(form)
 
+
+@APP.route('/rotationchanges')
```
bakert, line 256: I think you're right that this wants to be part of a "season summary" or similar eventually but this will do for a first stab.

decksite/templates/entry.mustache
```diff
@@ -1,5 +1,5 @@
 <li>
-    <a href="{{card.url}}" class="{{^card.pd_legal}}illegal{{/card.pd_legal}} card">{{n}} {{name}}</a>
+    <a href="{{card.url}}" class="{{^card.pd_legal}}illegal{{/card.pd_legal}} card">{{#n}}{{n}} {{/n}}{{name}}</a>
```
bakert, line 2: I would have anticipated you using singlecard for this not entry. But I see that singlecard doesn't have the bug details. I guess these two things are a bit too similar to make sense. I logged #3881 to clean this up.

decksite/views/rotation_changes.py
```diff
@@ -0,0 +1,14 @@
+from decksite.view import View
+
+# pylint: disable=no-self-use
+class RotationChanges(View):
+    def __init__(self, cards_in, cards_out):
+        self.sections = []
+        self.cards = cards_in + cards_out
+        entries_in = [{'n': None, 'name': c.name, 'card': c} for c in cards_in]
```
bakert, line 8: Do you need 'n': None here to avoid an exception when rendering?

decksite/views/rotation_changes.py
```diff
@@ -0,0 +1,14 @@
...
+        entries_out = [{'n': None, 'name': c.name, 'card': c} for c in cards_out]
+        self.sections.append({'name': 'New to the format', 'entries': entries_in, 'num_entries': len(entries_in)})
```
bakert, line 10: Nit. I'd probably use 'New to the Format' and 'Left the Format'.

decksite/views/rotation_changes.py
```diff
@@ -0,0 +1,14 @@
...
+        self.sections.append({'name': 'Left the format', 'entries': entries_out, 'num_entries': len(entries_out)})
```
bakert, line 11: A sweet additional feature would be "legal for the first time" and/or "only ever legal in this season".

magic/oracle.py
```diff
@@ -231,3 +231,24 @@ def insert_scryfall_card(sfcard):
...
+    previous = multiverse.SEASONS[multiverse.SEASONS.index(current_code) - 1]
+    previous_id = multiverse.get_format_id("Penny Dreadful {f}".format(f=previous))
+    current_id = multiverse.get_format_id("Penny Dreadful")
+    return changes_between_formats(previous_id, current_id)
+
+def changes_between_formats(f1, f2):
+    return [query_diff_formats(f2, f1), query_diff_formats(f1, f2)]
+
+def query_diff_formats(f1, f2):
+    query = 'c.id IN (SELECT card_id FROM card_legality WHERE format_id = {format1}) AND c.id NOT IN (SELECT card_id FROM card_legality WHERE format_id = {format2})'
```
bakert, line 246: Nit. I like to use triple-quoted strings to layout SQL is a more readable way. Not important.

magic/oracle.py
```diff
@@ -231,3 +231,24 @@ def insert_scryfall_card(sfcard):
...
+    removals_query = multiverse.base_query(where=query.format(format1=f1, format2=f2))
+
+    sql = """
```
bakert, line 249: This line seems redundant. If sql = base_query and base_query = removals_query then you only need one variable, not three? It was probably more complicated than this in some earlier version?

bakert (approved):
Awesome! A few tiny nits but nothing stopping it going to prod. Feel free to do a follow up PR if you think any of my nits are worth changing. Or not :)

decksite/views/rotation_changes.py
jgabrielygalan: I was pondering if it should say New this season. No idea of the equivalent for leaving.

bakert: 'New this Season' and 'Rotated Out'? What do they call it on dailymtg.com?

magic/oracle.py
silasary: Any reason you're using base_query and not cached_base_query?

bakert: Oh, that's important!

jgabrielygalan: No reason other than ignorance. I'll check that.

PR #3885, opened 2017-12-30
decksite: rotation changes fixes and nitpicks

jgabrielygalan:
Some fixes as commented in the other pull request:
- Use cached query
- Cleanup unnecessary variables
- Title text in the page
- Avoiding setting a dict key to None explicitly
- Import in alphabetical order


bakert (approved):
Awesome, thanks!

PR #3886, opened 2017-12-30
Private decks if current

jgabrielygalan:
Please review this and if possible test it locally yourself. This change is a big one, and very noticeable by people, so let's look at it with some more eyes.


bakert (approved):
Looks good to me and just in time for the new league :)

Regarding the copy I think we should avoid use of the word "user" and also avoid implying they've made a mistake so I'd replace "Logout and log in with the correct discord user" as just "Log out" probably.

bakert:
#movefast

bakert:
We should also send a 403 Forbidden from Download and Buy endpoints if visited for decks in current runs by unauthenticated users (I don't think there's any necessity to give them a flow to get logged in for those)

bakert:
Mouseover on the decklist still shows the full list so we'll need to block that too.

silasary:
Still not happy with the flow here.

It is completely possible for two users to sign up (unauthenticated), then view each others lists and accidentally associate with the wrong mtgo name.

decksite/main.py
```diff
@@ -32,9 +32,17 @@ def decks():
...
-@cached()
 def deck(deck_id):
-    view = Deck(ds.load_deck(deck_id))
+    d = ds.load_deck(deck_id)
+    person_from_discord = None
+    discord_user = session.get('id')
+    if discord_user is not None:
+        person_from_discord = ps.load_person_by_discord_id(discord_user)
+        if person_from_discord is None:
+            ps.associate(d, discord_user)
```
silasary, line 42: We need to get user confimation before doing something this permanent.

bakert:
This introduces state to the decklists too because of the mouseover thing. Ok we need a bit more thought here. I'm gonna revert for now and we can continue to work on it.

bakert:
Ok I'm not gonna revert it let's roll forward from here.

silasary:
Also of note, this makes it very hard to debug league stuff on local.

PR #3894, opened 2017-12-31
Hide tooltip decklists for active runs.

bakert:


silasary:
Manually merging, because Travis is having issues.

PR #3912, opened 2018-01-03
Private buy download

### LINKED ISSUE #3887: Restrict the download and buy server routes by discord user
If pull request https://github.com/PennyDreadfulMTG/Penny-Dreadful-Tools/pull/3886 is finally accepted, it's still missing the server side protection for the download and buy routes. What should be done: in those routes, if the deck is_in_current_run, then we should disallow the function if there's no logged user, or if the logged user is not the deck's owner.

jgabrielygalan:
Fixes #3887.

For the buy link, nothing needs to be done, since it's an external link which contains the decklist in the url, so as we are not generating the link it's not hackable.


decksite/auth.py
```diff
@@ -70,9 +72,31 @@ def decorated_function(*args, **kwargs):
...
+def logged_person():
+    return session.get('logged_person_id')
+
+def log_person(person_id):
+    session['logged_person_id'] = person_id
+
+def logged(f):
+    @wraps(f)
+    def decorated_function(*args, **kwargs):
+        id = discord_id()
```
silasary: Rename this variable.  `id` is the name of a built-in function, and shouldn't be used as a variable name.

silasary: This is why Codacy is failing.

silasary (changes requested):

silasary (approved):

Issue #1546, opened 2017-11-06
Create some kind of schedule to update and test versioned libraries


bakert:
Python, MySQL, jQuery, everything in footer.mustache, everything in requirements.txt can all fix security holes, become faster, get better, etc. And big migrations are much harder than small ones. We should make a schedule to regularly update to latest stable versions and test.

bakert:
We should also fix our versions in requirements.txt with == to avoid usual devs having working envs but everything exploding when a new dev tries to set up.

silasary:
In addition to the above commit, I've set up requires.io to automatically monitor the dependencies, and create PRs when any of them updates.

As for mysql, python, and other distro packages, I already do a full `zypper update` once per month (usually on the 30th).

The server runs OpenSUSE Tumbleweed (rolling release), so we don't need to worry about doing dist-upgrades or any of that silliness.

bakert:
This is awesome :D

PR #3921, opened 2018-01-04
Freeze requirements as per #1546

silasary:


silasary:
Also worth noting:  As part of updating requirements, I dug into the issues I was having installing `googler` (or more specifically, its dependency `pycrypto`).

See https://github.com/PennyDreadfulMTG/googler/commit/ccc5e2b7703678583f389bfc3199d7469a692a74 for more details.

PR #3942, opened 2018-01-04
Rotation speculation

jgabrielygalan:
In order to merge this PR, we need to apply these indexes in prod:

create INDEX idx_name_name on magic._cache_card(name(142))
create INDEX idx_name_price on prices.cache(price)


decksite/main.py
```diff
@@ -275,11 +275,16 @@ def do_claim():
     return retire(form)
 
 
-@APP.route('/rotationchanges')
+@APP.route('/rotationchanges/')
```
bakert: Let's make this /rotation/changes/ and /rotation/speculation/?

PR #4000, opened 2018-01-10
Revert "Switch to Scryfall's CDN for card images."

bakert:
This reverts commit 5b8a968db46a3040b93c45db8c176607d27f9a28.

I've changed bluebones back to how it was but with a special case for BFM.

Scryfall don't offer a simple way to get the reverse of meld cards.


silasary:
Sounds good

silasary:
Bors r+

PR #4099, opened 2018-01-19
Index fixes + automatic disambiguation

jgabrielygalan:


requirements.txt
```diff
@@ -1,5 +1,4 @@
 anytree==2.4.3
-astroid==1.5.3
```
silasary: ??

jgabrielygalan: I was trying to force the astroid version due to this bug: https://github.com/PyCQA/pylint/issues/1784 but it didn't help, so I removed it

PR #4139, opened 2018-01-24
Protect against unknown end_date when loading season info.

bakert:


bakert:
Dear Travis,

This is really, really, really fun but please do consider not failing at an arbitrary point during installing requirements with no error message this time. Thanks! Pip pip Cheerio!

Tom

PR #4221, opened 2018-02-04
Reconnect

### LINKED ISSUE #4213: Add reconnection capability to database_mysql
If the connection is dropped for some reason (wait_timeout or whatever), the bot doesn't know how to reconnect.

jgabrielygalan:
Fixes #4213 

Hopefully makes Tom happier without all those "MySQL has gone away" errors :)

### DIFF
--- a/shared/database_mysql.py
+++ b/shared/database_mysql.py
@@ -58,7 +58,7 @@ class MysqlDatabase(GenericDatabase):
 
     def execute_with_reconnect(self, sql, args):
         # Attemp to excute the query and reconnect 3 times, then give up
-        for attempt in range(3):
+        for _ in range(3):
             try:
                 p = perf.start()
                 self.cursor.execute(sql, args)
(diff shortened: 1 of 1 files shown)


shared/database_mysql.py
```diff
@@ -55,6 +56,28 @@ def execute(self, sql, args=None):
...
 
+    def execute_with_reconnect(self, sql, args):
+        result = None
+        # Attempt to excute the query and reconnect 3 times, then give up
+        for _ in range(3):
+            try:
+                p = perf.start()
+                self.cursor.execute(sql, args)
+                perf.check(p, 'slow_query', (sql, args), 'mysql')
+                result = self.cursor.fetchall()
```
silasary, line 67: Just return the result here if it works.

jgabrielygalan: It works but codacy complains about a for with else and no break. And with the break, it complains about no reachable code, so I left it like this, with the result variable to comply with both rules.

bakert:
Yayyyyyyyuyh

Issue #4224, opened 2018-02-05
Bugs wording and consistency


bakert:
On /tournaments/ under Bugs, "Cards with game-breaking bugs should not be played" is ambiguous. It should read "may not be played" and the current cards with game breaking bugs in common play should be added as exceptions or have their label changed. Many of the unplayed gamebreaking bug cards clearly can never be allowed, but the two that are in heavy play seem fine at the players own risk.

silasary:
According to RFC 2119, `SHOULD NOT` is stronger than `MAY NOT`.

bakert:
That is an interesting fact. @mmrsad – are we done here?

PR #4228, opened 2018-02-06
Fix /tournaments/ bug rules

mmrsad:
Update /tournaments/ rules with "must not" (the strongest verbiage) and corrected exception list to 1 card per stash86, Sylvan Primordial

https://github.com/PennyDreadfulMTG/Penny-Dreadful-Tools/issues/4224


silasary:
Good enough for now.
Ideally we dynamically generate the exceptions in the future.

PR #4229, opened 2018-02-06
Barebones l18n support

silasary:
Which is to say, the localization stuff is there.  I just haven't mapped many strings.


silasary:
Ready to merge

PR #4266, opened 2018-02-10
minor fixes

silasary:


bakert:
It’s probably preferable to accept markdown’s views rather than tampering
with it. If you want <br> instead there are ways

PR #4345, opened 2018-02-24
Dismissable intro box to help orient new players.

bakert:


decksite/views/about.py
```diff
@@ -50,12 +50,11 @@ def fancy_cards():
         'Dragonstorm',
         'Day of Judgment',
         'Cruel Ultimatum',
-        'Rofellos, Llanowar Emissary',
```
silasary, line 53: No love for Rofellos?

bakert: Because of the way we write these out it just reads weird when the cards have a comma in the name:

"Treasure Cruise, Rofellos, Llanowar Emissary and Day of Judgment"

we could use semicolons but that's pretty ugly.

silasary: The way that sentence is written, it's currently not localizable.  So we may want to rethink it anyway.

But for now, I'm good with removing him from the list.

bakert: It could be a bulleted list but if you're still determined to achieve fully-functioning i18n we're going to have to hand translators complete sentences (always) with placeholders. There's no other way.

And it still won't work because of "Adam's Page". (In French that's "Page de Jean" but "Page d'Anita" -- have fun getting your translators do that translation …) And let's not get started on parts of speech that change based on number is involved. I think it's Polish that particularly cares about whether a number is greater than or less than 200 …

In case it's not clear I'm down on i18n :D I'll try not to make it harder/worse but I really don't want to expend effort on making it good. Tar pit warning!

PR #4385, opened 2018-03-02
Link to rulings more specifically as sugested by scryfall folks.

### LINKED ISSUE #4359: !rulings can use the #rulings anchor on Scryfall
I noticed that if there are too many rulings to display for a card, the MTGBot will link to Scryfall. You can add the `#rulings` anchor to jump people directly to that portion of the page when they open the link.

Ex:
https://scryfall.com/card/cn2/78
vs 
https://scryfall.com/card/cn2/78#rulings

We use this ID anchor for our own bot, so it won't suddenly change or be removed

45ywvzl1s1

bakert:
Fixes #4359.

### DIFF
--- a/discordbot/command.py
+++ b/discordbot/command.py
@@ -707,7 +707,7 @@ def card_rulings(c):
     if len(rulings) > 3:
         n = len(rulings) - 2
         rulings = rulings[:2]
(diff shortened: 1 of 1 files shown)


bakert:
This was done by @jgabrielygalan i think.

PR #4582, opened 2018-04-14
Cleanup

silasary:


magic/rotation.py
```diff
@@ -53,4 +53,8 @@ def text():
         return "The supplemental rotation is in {sdiff} (The next full rotation is in {diff})".format(diff=dtutil.display_time(diff.total_seconds()), sdiff=dtutil.display_time(sdiff.total_seconds()))
     return "The next rotation is in {diff}".format(diff=dtutil.display_time(diff.total_seconds()))
 
-SETS = init()
+__SETS = []
+def sets():
+    if __SETS is None:
```
bakert: Will __SETS ever be None if we've initialized it to `[]`?

PR #4595, opened 2018-04-17
Minor typechecked config improvements

bakert:
This will help us find the offender faster.


bakert:
I regret everything. I don't think mypy is going to be smart enough to follow this. Which is probably why Katelyn wrote it the other way in the first place.

bakert:
Abandon ship!

Issue #4583, opened 2018-04-14
Match URL not just text for resources search


vorpal-buildbot:
Reported on Discord by bakert#2193

bakert:
For example "!resources github" should find any entry with "github" in the url as well as in the name.

PR #4607, opened 2018-04-18
make resources match on url + isort changes

jgabrielygalan:
Fixes: #4583


bakert:
That is a lot of isort chagnes. I assume yours is somehow configured
differently rather than that these changes are actually wanted but
@silasary would know for sure.

2018-04-18 21:29 GMT+02:00 jgabrielygalan <notifications@github.com>:

silasary:
Yeah, considering that isort is failing in on the travis build, it looks like isort is either picking up the wrong settings from somewhere, not picking up your virtualenv, or not up to date.

jgabrielygalan:
No idea, I have: 

isort (4.3.4)

and: 

find . -name '.isort.cfg'
./decksite/charts/.isort.cfg

I could try to skip the isort changes and upload only my own changes to push this through, but I woud like to be aligned with you on the tools.

silasary:
Not sure why your machine is getting confused.  

It looks like isort doesn't recognise the difference between 3rd-party and local packages.  Which either means it's not detecting your virtualenv, or you haven't installed all the requirements?

You're on the correct version.

PR #4612, opened 2018-04-19
Remove stdout-fritzing. Seems like prod is able to cope with unicode now.

bakert:


bakert:
I am guessing that it *wasn't* safe to remove the sys.stdout stuff. Hmmmm.

PR #4627, opened 2018-04-20
Price grabber was doing bad things.

silasary:
I have a feeling that a SQL query that inserts 16,601 rows in a single INSERT might be bad for mysql.

bors r+


bakert:
Did you time it? My guess is that this is going to be slower and that it's the (single) caching query that's still the slowest thing. But I live to be wrong :D

silasary:
Yeah, this was significantly faster for me on local.

The mysql buffer was bottlenecking on me.

silasary:
The cache query all happens within mysql, and seems to be pretty good at what it does.

PR #4641, opened 2018-04-21
Kill season column. We have season pages now.

bakert:


silasary:
Seasons column is valuable for several reasons, and shouldn't be killed.
image
image

silasary (changes requested):
As mentioned on Discord, I am strongly opposed to these changes.

PR #4684, opened 2018-04-24
Fix all extant typing issues and make typing required in make lint.

bakert:


bakert:
Sweet, so:

- Typing works for mypy.
- Lint works for pylint.
- Tests work for `make unit`.
- Imports work for isort.

but I can't land because however jenkins accumulates tests to run doesn't like the type annotations.

Marvellous :D

:(

bakert:
bors r+

i believe

PR #4719, opened 2018-04-29

silasary:


silasary:
Um?

bakert:
I think one of the feeds changed. Prices are not updating because of this same thing.

bakert:
actually `2015-10-02T00:00:00.000` is a pure bs date so one of the feeds changed AND it's buggy.

silasary:
Oh.  It's almost certainly https://github.com/glacials/whatsinstandard/commit/606e2fbbac7275be984aabb273e0556f643dc60e

bakert:
ah, good find. "nonbreaking change" ho ho ho. these people have got too used to php's strtotime :D

PR #4731, opened 2018-05-02

silasary:


bakert:
bors r+

2018-05-02 19:00 GMT+01:00 Katelyn Gigante <notifications@github.com>:

silasary:
From the MyPy 0.600 changelog:
* Enable strict optional checking by default.

silasary:
We have two options. 
Disable that option, or mark things as optional where it's warning us.

PR #4737, opened 2018-05-04
mypy 0.600 compatibility

silasary:


silasary:
@bakert I got all but one of them, but I can't figure out what this last error is.

Can you investigate?

bakert:
I think this might be a bug in 0.600 but changing the Nones in the tuple we pass in to empty strings convinces it that everything is ok so doing that. This ParseResult is a sublcass of Tuple AnyStr is a generic for [bytes, str] nonsense all colliding boils my brain. But this works :D

PR #4740, opened 2018-05-04
Homepage perf improvement, create perf regression test.

bakert:


bakert:
oh, ok bors :D you do you

PR #4757, opened 2018-05-08
Logsite+

silasary:
* Logsite tests
* Subclass Flask, and move common code into it.
* Actual better error pages. (+Card images for logsite, -db reliance for decksite)


silasary:
bors try

PR #4849, opened 2018-05-23
Make all tests and lint pass, as it should be. Peace.

bakert:


decksite/templates/editmatches.mustache
```diff
@@ -7,6 +7,9 @@
                     <td>
                         {{display_date}}
                     </td>
+                    <td>
+                        {{mtgo_id}}
```
silasary: This is a duplicate of lines 27-32

silasary (changes requested):

decksite/templates/editmatches.mustache
bakert: Oh, I guess we already did this :D

silasary: The ticket was that that column didn't actually work properly 😜

silasary:
Don't revert it entirely!  The SQL addtion is still needed.

silasary (approved):

Issue #1963, opened 2017-11-19
Set Expires headers on our CSS and js


bakert:

bakert:
This is in the webserver config not Flask I think.

We can set it into the far future because our static resources have the current commit id on them. (We should possibly get a bit more sophisticated than that and only use the id of the last commit that modified them.)

bakert:
Looks like we get 12 hours as currently configured:

        > GET /static/css/pd.css HTTP/1.1
        > Host: pennydreadfulmagic.com
        > User-Agent: curl/7.55.0
        > Accept: */*
        >
        < HTTP/1.1 200 OK
        < Date: Mon, 20 Nov 2017 07:03:06 GMT
        < Content-Type: text/css; charset=utf-8
        < Content-Length: 298089
        < Connection: keep-alive
        < Set-Cookie: __cfduid=dc68008e197ab6b9010c2c51969b0e0d31511161386; expires=Tue, 20-Nov-18 07:03:06 GMT; path=/; domain=.pennydreadfulmagic.com; HttpOnly; Secure
        < Last-Modified: Wed, 15 Nov 2017 06:41:35 GMT
        < Cache-Control: public, max-age=43200
        < Expires: Mon, 20 Nov 2017 19:03:06 GMT
        < ETag: "1510728095.200065-298089-3468169871"
        < CF-Cache-Status: REVALIDATED
        < Accept-Ranges: bytes
        < Server: cloudflare-nginx
        < CF-RAY: 3c098c28b9832a1f-SEA

bakert:
<img width="714" alt="captura de pantalla 2017-11-25 a las 7 42 49 p m" src="https://user-images.githubusercontent.com/1541/33236940-e3bca4aa-d218-11e7-8719-ccbf6f1e9f36.png">

silasary:
Told Cloudflare to bump it to a month.

silasary:
We definitely need to use file hashes, rather than the repo hash though.  Otherwise it'll never last that long.

bakert:
d51c137d uses last commit *of the file*.

PR #4868, opened 2018-05-29
Fix regression introduced in d51c137d79c723ecd2e0d408fd052aca8f070837.

silasary:
The fix for #1963 involved running three seperate `git` executions per page request.
This commit fixes that by once again caching commit-ids on the flask application instance.

bors r+

### DIFF
--- a/logsite/view.py
+++ b/logsite/view.py
@@ -2,9 +2,6 @@ from flask import url_for
 
 from shared_web.base_view import BaseView
 
-from . import APP
-
-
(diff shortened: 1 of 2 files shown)


silasary:
bors retry

bakert:
Did it matter? There are some things we do hundreds of thousands of times
per request :)

silasary:
There are.  But executing subprocesses are more expensive than most other things we do.

PR #5065, opened 2018-07-19
Remove failed experiment in cache updating.

bakert:
This caused more perf issues hitting the site 20+ every four minutes than
it solved keeping the cache hot.


silasary:
Merge Conflict

bakert:
Interesting. There is no merge conflict.

bakert:
Creating a new branch from master because I can't get this one to see sense.

PR #5087, opened 2018-07-28
An experiment

silasary:
Not sure whether I like this or not.

Please give opinions when it deploys to beta site.


silasary:
We need banners for: 

- [x] Season 1
- [x] Season 2
- [x] Season 3
- [x] Season 4
- [x] Season 5
- [x] Season 6
- [ ] "All Seasons"

- [ ] And maybe a less busy S9?

bakert:
I'm opposed to anything that makes the site slower until the site is super fast.

It's kinda nice but that's the tiebreaker for me. Shall we kill this PR for now?

PR #5124, opened 2018-08-09
Make Deck less meaty.

silasary:


bakert:
CardRef is a surprisingly awkward thing to name. "Slot"? That doesn't do it for me either.

silasary:
`DecklistLine`?

bakert:
DeckEntry, DeckCard. Hm. None of these are any better than CardRef. I'll mull.

PR #5173, opened 2018-08-21
News for homepage - tournament wins, league 5-0s and code merges.

bakert:


bakert:
False positive from Codacy, have forgotten how to force this through, if I ever knew.

PR #5186, opened 2018-08-24
don't query github every time we load '/'

silasary:


bakert:
We should probably cache all of news for at least an hour if not longer. It's moderately expensive to calculate the other stuff too, I think.

PR #5240, opened 2018-08-31
Hide title of active league run in a couple of places it was leaking.

bakert:


discordbot/command.py
```diff
@@ -784,7 +784,7 @@ def more_results_link(args: str, total: int) -> str:
         await post_no_cards(channel, replying_to)
         return
     not_pd = configuration.get_list('not_pd')
-    disable_emoji = str(channel.id) in not_pd or str(channel.guild.id) in not_pd
+    disable_emoji = str(channel_id(channel)) in not_pd
```
silasary, line 787: Um, this is not what this does.

bakert: Hm, I see, we need to check both not just one as it can be disabled per channel or per server?

PR #5279, opened 2018-09-05
Fix !art command

silasary:


bakert:
Killed in favor of #5287.

silasary:
We do need to make this refactor in order to support using non cropped images on decksite. But yours is good for the time being

bakert:
I did that with CSS instead. Check it out!

PR #5331, opened 2018-09-12
Make archetypes query zillions of times faster.

bakert:


bakert:
Ah, yeah, this will not bootstrap.

PR #5420, opened 2018-09-22

silasary:


silasary:
Breaking changes upstream.

bakert:
There is a Markdown 3.0.1 now so I wonder if this can go through.

bakert:
Reverted the _ change to please codacy and see if this goes through now …

bakert:
ok I get it now. Codacy doesn't want us to have an unused param that doesn't start with '_' and pylint doesn't want us to override a param without using the same name as the param it overrides. I would say pylint is "wrong" here although it's a marginal call between two broadly sensible ideas. I'll just exclude it from pylint inline.

PR #5473, opened 2018-10-03
Cache rotation scores in redis for perf.

bakert:


bakert:
Abandoning this because it's a partial solution and the typing is too difficult.

PR #5555, opened 2018-10-15
Correctly add all meld cards to mtgbot's search index

### LINKED ISSUE #5382: mtgbot can't find cards with the bottom half of a meld card on the back (eg midnight scavengers)
Reported on Discord by acidicrhyme#5579

adambiltcliffe:
Previously meld cards with position=2 were not being added correctly but were indexed under each individual letter of their name instead due to a subtle bug in whoosh_write

Fixes #5382


bakert:
Good spot. Thanks!

PR #5582, opened 2018-10-18
Include jquery in the jquery-approved way.

bakert:


bakert:
This actually messes up table rendering moving us to latest jquery so we'll have to look into it.

PR #5599, opened 2018-10-20
Add internal archetype classification guidance for admins

### LINKED ISSUE #5492: Allow admins to add/edit an annotation to each archetype for showing on edit archetypes page

adambiltcliffe:
Interface could be jazzier, but gradually populating the db with some guidance about what the archetypes actually mean might make life easier for new demimods like me. Closes #5492


silasary (changes requested):
Can you call it `description` instead?

adambiltcliffe:
I deliberately avoided using `description` to make it clear that it's to inform classification only and never intended to be visible to non-admin users. Can change if you prefer though.

adambiltcliffe:
Now renamed

Issue #5604, opened 2018-10-21
Add 'generalist' and 'specialist' achievements for winning/top-8ing a tournament with X different archetypes in the same season and with the same archetype Y times


vorpal-buildbot:
Reported on Discord by acidicrhyme#5579

adambiltcliffe:
#5629

PR #5616, opened 2018-10-22
Achievement updates and new Generalist achievement

### LINKED ISSUE #5523: Include count of seasons in /seasons/all on completionist badge (but not on each season as that's boolean)

### LINKED ISSUE #5600: Refactor the way that achievements are handled so that adding new ones doesn't require modifying the template or view
Reported on Discord by acidicrhyme#5579

adambiltcliffe:
- Big internal restructure of achievements. All data and logic related to each individual achievement is now contained in a single class (closes #5600)
- Once-per-season achievements such as Completionist now show the number of seasons you earned them in on your achievements page if seasons is set to "all" (closes #5523)
- New Generalist achievement for top-8ing three tournaments in a season with three different decks (half of #5604)


decksite/achievements.py
```diff
@@ -0,0 +1,237 @@
...
+from flask import url_for
+from flask_babel import ngettext
+
+import decksite
+from decksite.data import query
+from magic import tournaments
+
+if TYPE_CHECKING:
+    from decksite.data import person # pylint:disable=unused-import
+# Disabling unused-import supposedly not needed here but actually seems to be?
```
bakert, line 12: nit. Prefer tasks to TODOs.

decksite/achievements.py
```diff
@@ -0,0 +1,237 @@
...
+
+def load_query(people_by_id: Dict[int, 'person.Person'], season_id: Optional[int]) -> str:
+    columns = ', '.join(f'SUM({a.key}) as {a.key}' for a in Achievement.all_achs if a.in_db)
```
bakert, line 15: nit. It wouldn't kill you to say all_achievements :D thought you were referring to archetypes for a second.

decksite/achievements.py
```diff
@@ -0,0 +1,237 @@
...
+        FROM
+            _achievements AS a
+        WHERE
+            person_id IN ({ids}) AND ({season_query})
+        GROUP BY
+            person_id
+    """.format(columns=columns, ids=', '.join(str(k) for k in people_by_id.keys()), season_query=query.season_query(season_id))
+
+def preaggregate_query() -> str:
+    create_columns = ', '.join(f'{a.key} INT NOT NULL' for a in Achievement.all_achs if a.in_db)
```
bakert, line 29: We should probably backtick-escape (in case we use a SQL reserved word) and/or sql-escape (for safety) these when we interpolate them into the string unless we're like super duper sure that's unnecessary and no one will ever fuck that up in the future.

decksite/achievements.py
```diff
@@ -0,0 +1,237 @@
...
+
+def descriptions() -> List[Dict[str, str]]:
+    return [{'title': a.title, 'description_safe': a.description_safe} for a in Achievement.all_achs]
+
+def displayed_achievements(p: 'person.Person') -> List[Dict[str, str]]:
+    return [d for d in (a.display(p) for a in Achievement.all_achs) if d is not None]
+
+# Abstract achievement classes
+
+class Achievement:
```
bakert, line 67: Cool stuff. I had not seen Python classes used like this before (like a struct, and with @property). I wonder if some of our models should use this style instead of being Containers.

bakert (commented):
Pretty sweet. A couple of nits but nothing that should stop us having this.

decksite/achievements.py
silasary: I've been morphing a couple of our Containers this way.

PR #5629, opened 2018-10-24
New achievement and simple global achievement stats

adambiltcliffe:
New Specialist achievement for making the knockout rounds of 3 tournaments in a season playing the same archetype
Achievements page now shows how many players have earned each achievement

### DIFF
--- a/decksite/achievements.py
+++ b/decksite/achievements.py
@@ -5,6 +5,8 @@ from flask import url_for
 from flask_babel import ngettext
 
 import decksite
+from shared.container import Container
+from decksite.database import db
 from decksite.data import query
 from magic import tournaments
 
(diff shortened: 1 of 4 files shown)


decksite/achievements.py
```diff
@@ -91,6 +93,12 @@ def display(self, p):
         if n > 0:
             return {'name': self.title, 'detail': ngettext(f'1 {self.singular}', f'%(num)d {self.plural}', n)}
         return None
+    def load_summary(self) -> Optional[str]:
+        sql = f"""select sum(`{self.key}`) as count, sum(case when `{self.key}` > 0 then 1 else 0 end) as pcount from _achievements"""
```
bakert: nit. Please stick to `SELECT SUM(), SUM(CASE WHEN blah > 0 THEN 1 ELSE 0 END) AS pcount FROM ...` style for consistency. 

You might want to avoid using count as a var name as it's a sql reserved word or put it in backticks like we do with the match table.

bakert (commented):
I'm not sure I totally see where this is going but nothing unreasonable here to my eye.

PR #5637, opened 2018-10-25
Speed up loading the cards for decks.

bakert:


silasary:
It exists, it's just brand new (released two hours ago).  Might not have propagated across all the PyPy mirrors yet.

PR #5656, opened 2018-10-25
Factor out some repeated code

adambiltcliffe:


silasary:
The name feels bad and unobvious. 
But I like what it does

adambiltcliffe:
I couldn't think of a better name that wasn't really long. `on_error_call_func_and_retry`?

PR #5660, opened 2018-10-26
Percentage bars

silasary:


decksite/achievements.py
```diff
@@ -94,7 +94,6 @@ def display(self, p: 'person.Person') -> str:  # pylint: disable=no-self-use, un
     def load_summary(self, season_id: Optional[int] = None) -> Optional[str]:
         season_condition = f'AND season_id = {season_id}' if season_id != 'all' else ''
         sql = f"""SELECT SUM(`{self.key}`) AS num, COUNT(DISTINCT person_id) AS pnum FROM _achievements WHERE `{self.key}` > 0 {season_condition}"""
-        print(sql)
```
silasary, line 97: @adambiltcliffe Please don't use print() on decksite.  It's really bad for performance.

decksite/main.py
```diff
@@ -94,6 +94,10 @@ def achievements():
     view = Achievements(auth.mtgo_username())
     return view.page()
 
+@APP.route('/person/achievements/')
```
silasary, line 97: @adambiltcliffe Also try not to break old URLs.  
If you change any other URLs in the future, please add a redirect.  (It's bad for SEO, old links, and people's bookmarks)

decksite/achievements.py
```diff
@@ -103,6 +102,14 @@ def load_summary(self, season_id: Optional[int] = None) -> Optional[str]:
             players_text = ngettext('1 player', f'%(num)d players', res.pnum)
             return f'Earned{times_text} by {players_text}.'
         return None
+    def percent(self, season_id: Optional[int] = None) -> float:
+        season_condition = f'season_id = {season_id}' if season_id != 'all' else ''
```
bakert, line 106: This was the 500 on /seasons/all/achievements – we have `query.season_query(season_id)` to encapsulate this logic.

decksite/achievements.py
```diff
@@ -103,6 +102,14 @@ def load_summary(self, season_id: Optional[int] = None) -> Optional[str]:
...
+        sql = f"""SELECT SUM(CASE WHEN {self.key} > 0 THEN 1 ELSE 0 END) AS pnum, COUNT(*) AS mnum FROM _achievements WHERE {season_condition}"""
```
bakert, line 107: nit. Triple quotes for multiline strings, single quotes for single-line strings, double quotes for single-line strings containing apostrophes.

decksite/templates/achievements.mustache
```diff
@@ -4,7 +4,12 @@
         {{#achievement_descriptions}}
         <dt class='{{class}}'>{{title}}</dt>
         <dd>{{{description_safe}}}</dd>
-        {{#summary}}<dd>{{summary}}</dd>{{/summary}}
+        {{#summary}}
+            <dd class='achievement_holder'>
```
bakert, line 8: nit. Double quotes for html attribute values.

bakert (commented):
This is cool! I think we need to make it a bit clearer what it is but good stake in the ground.

decksite/achievements.py
silasary: This was me blindly copying the code above after merging with head at midnight

bakert: c8e43f7d fixes the above call.

Issue #5657, opened 2018-10-25
`!random deck` gives you a random legal deck


vorpal-buildbot:
Reported on Discord by bakert#2193

adambiltcliffe:
Should this be a random deck from the current season, or do we want to include ones from previous seasons?

bakert:
the request was for current season

adambiltcliffe:
Currently mtgbot doesn't touch the decksite database (does it?). Does this open any potential cans of worms?

bakert:
mtgbot talks to decksite via the API reading (for example) /api/sitemap. So exposing it in a similar way would make sense. Although having an API endpoint to return all decks is madness and a random deck is weird. But so be it! Or the bot could get max id then generate a random one then check the page isn't a 404. Not sure if that's worse or better!

bakert:
(Mostly it's not API stuff – mtgbot just "knows" a few things about the site probably slightly in appropriately but certainly doesn't delve into its innards.)

silasary:
I'm ok with an `/api/randomlegaldeck` endpoint

PR #5668, opened 2018-10-27
add API endpoints for getting a random legal deck

adambiltcliffe:
/api/randomlegaldeckurl/ returns a JSON object with a string property 'url' that gives the page for a random deck from the current season. For use in #5657

### DIFF
--- a/decksite/api.py
+++ b/decksite/api.py
@@ -18,6 +18,20 @@ def deck_api(deck_id):
     blob = deck.load_deck(deck_id)
     return return_json(blob)
 
+@APP.route('/api/randomlegaldeck/')
+def random_deck_api():
+    blob = deck.random_legal_deck()
+    return return_json(blob)
+
+@APP.route('/api/randomlegaldeckurl/')
+def random_deck_url_api():
+    d = deck.random_legal_deck()
+    if d is None:
+        result = {'url': None, 'error':True, 'message':'No legal decks could be found'}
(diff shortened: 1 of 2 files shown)


decksite/api.py
```diff
@@ -18,6 +18,20 @@ def deck_api(deck_id):
     blob = deck.load_deck(deck_id)
     return return_json(blob)
 
+@APP.route('/api/randomlegaldeck/')
```
silasary: No trailing slash

decksite/api.py
```diff
@@ -18,6 +18,20 @@ def deck_api(deck_id):
...
+def random_deck_api():
+    blob = deck.random_legal_deck()
+    return return_json(blob)
+
+@APP.route('/api/randomlegaldeckurl/')
```
silasary: This seems unneccessary

silasary: Just use the above endpoint and extract the url for it

silasary (changes requested):

decksite/api.py
adambiltcliffe: Removed

decksite/api.py
adambiltcliffe: Removed /api/randomlegaldeckurl and added url attribute to the returned data from /api/randomlegaldeck

silasary (approved):

PR #5675, opened 2018-10-28
Add confidence measure to archetype predictions

adambiltcliffe:
Add confidence measure to archetype predictions.

Tagged this with DNM because it modifies the database but feel free to remove if it looks sane.

Also may be worth running the hourly maintenance tasks manually once this goes live since it will need to re-predict all the ones currently in the queue.

### DIFF
--- a/decksite/data/archetype.py
+++ b/decksite/data/archetype.py
@@ -121,9 +121,14 @@ def add(name: str, parent: int) -> None:
     sql += '({ancestor}, {descendant}, {depth})'.format(ancestor=archetype_id, descendant=archetype_id, depth=0)
     db().execute(sql)
 
-def assign(deck_id: int, archetype_id: int, reviewed: bool = True) -> None:
+def assign(deck_id: int, archetype_id: int, reviewed: bool = True, confidence: Optional[int] = None) -> None:
     and_clause = '' if reviewed else 'AND reviewed is FALSE'
     db().execute(f'UPDATE deck SET reviewed = %s, archetype_id = %s WHERE id = %s {and_clause}', [reviewed, archetype_id, deck_id])
+    if not reviewed and confidence is not None:
+        db().execute(f'INSERT INTO confidence VALUES (%s, %s) ON DUPLICATE KEY UPDATE score=%s', [deck_id, confidence, confidence])
+
+def clear_old_predictions() -> None:
+    db().execute('DELETE confidence FROM confidence LEFT JOIN deck ON confidence.deck_id = deck.id WHERE deck.reviewed = TRUE')
 
 def load_all_matchups(where: str = 'TRUE', season_id: Optional[int] = None, retry: bool = False) -> List[Container]:
     sql = """
--- a/decksite/data/archetype.py
+++ b/decksite/data/archetype.py
@@ -121,14 +121,11 @@ def add(name: str, parent: int) -> None:
     sql += '({ancestor}, {descendant}, {depth})'.format(ancestor=archetype_id, descendant=archetype_id, depth=0)
     db().execute(sql)
 
-def assign(deck_id: int, archetype_id: int, reviewed: bool = True, confidence: Optional[int] = None) -> None:
+def assign(deck_id: int, archetype_id: int, reviewed: bool = True, similarity: Optional[int] = None) -> None:
     and_clause = '' if reviewed else 'AND reviewed is FALSE'
     db().execute(f'UPDATE deck SET reviewed = %s, archetype_id = %s WHERE id = %s {and_clause}', [reviewed, archetype_id, deck_id])
-    if not reviewed and confidence is not None:
-        db().execute(f'INSERT INTO confidence VALUES (%s, %s) ON DUPLICATE KEY UPDATE score = %s', [deck_id, confidence, confidence])
-
-def clear_old_predictions() -> None:
-    db().execute('DELETE confidence FROM confidence LEFT JOIN deck ON confidence.deck_id = deck.id WHERE deck.reviewed = TRUE')
+    if not reviewed and similarity is not None:
+        db().execute(f'UPDATE deck_cache SET similarity = %s WHERE deck_id = %s', [similarity, deck_id])
 
 def load_all_matchups(where: str = 'TRUE', season_id: Optional[int] = None, retry: bool = False) -> List[Container]:
     sql = """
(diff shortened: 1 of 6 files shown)


decksite/data/archetype.py
```diff
@@ -121,9 +121,14 @@ def add(name: str, parent: int) -> None:
     sql += '({ancestor}, {descendant}, {depth})'.format(ancestor=archetype_id, descendant=archetype_id, depth=0)
     db().execute(sql)
 
-def assign(deck_id: int, archetype_id: int, reviewed: bool = True) -> None:
+def assign(deck_id: int, archetype_id: int, reviewed: bool = True, confidence: Optional[int] = None) -> None:
     and_clause = '' if reviewed else 'AND reviewed is FALSE'
     db().execute(f'UPDATE deck SET reviewed = %s, archetype_id = %s WHERE id = %s {and_clause}', [reviewed, archetype_id, deck_id])
+    if not reviewed and confidence is not None:
+        db().execute(f'INSERT INTO confidence VALUES (%s, %s) ON DUPLICATE KEY UPDATE score=%s', [deck_id, confidence, confidence])
```
bakert: nit. `score = %s` (spacing).

decksite/data/deck.py
```diff
@@ -502,6 +502,15 @@ def load_conflicted_decks() -> List[Deck]:
...
     return load_decks(where, order_by='d.decklist_hash')
 
+def load_confidence(decks: List[Deck]) -> None:
+    # Confidence scores for decks not in the queue get deleted by the hourly maintenance job, so just get all
+    sql = 'SELECT deck_id, score FROM confidence'
+    conf = {}
+    for row in (Container(r) for r in db().select(sql)):
+        conf[row.deck_id] = row.score
+    for deck in decks:
+        deck.confidence = f'{conf[deck.id]}%' if deck.id in conf else ''
```
bakert: This is our first (?) use of more than a simple variable in `{}` inside a string. I'm not sure how I feel about it. f''-strings are super useful but rather like mustache templates I'd hate to see them getting complicated. Probably not a big deal.

decksite/sql/46.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS confidence (
+    deck_id INTEGER PRIMARY KEY UNIQUE,
+    score INTEGER,
+    FOREIGN KEY(deck_id) REFERENCES deck(id) ON UPDATE CASCADE ON DELETE CASCADE
+);
+
+-- Clear existing predictions so that we recalculate a confidence score for them
+UPDATE deck SET archetype_id=NULL WHERE reviewed=FALSE;
```
bakert: nit. `archetype_id = NULL WHERE reviewed = FALSE;` (spacing)

decksite/sql/46.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS confidence (
```
bakert: Alternative names for this table – deck_confidence or _deck_confidence.

There's a little stylistic confusion about what we call tables and it's my fault.

I think we should use the following naming scheme:

foo – In the singular, a table that contains a row per foo (person, deck).
foo_bar – The conjunction of a foo and a bar and associated data (deck_match, deck_card).
_foo_stats, _baz – Tables that are deleted and recreated in their entirety from values elsewhere during preaggregation. No original data lives in these tables.

Which leaves us with just `deck_cache` and `confidence` as things that don't fit into any of these buckets. deck_cache is "things that are derivable from other things in the db + code logic but we tear it down and rebuild it deck-by-deck during reprime_cache and omw gets updated separately" and confidence is remarkably similar to that. There's no underivable data in confidence you just apply code logic to the existing database and out comes confidence. That makes it more like the _ tables with some differences in how it is formed than like the actual data tables.

Options:

1. Rename deck_cache to _deck_cache. Call confidence _deck_confidence. Ignore how these tables are formed and just use the _ to indicate "if you deleted this table and there were no backups we wouldn't care".
2. Leave deck_cache as is and rename confidence to confidence_cache.
3. Leave deck_cache as is and rename confidence to deck_confidence_cache.

Separately thinking about all this leads me to the question:

Why not put the confidence column into deck_cache?

It is a derivable thing about a deck generated by a job that can change going forward just like omw or wins or the sanitized name. It seems like a pretty perfect fit now that I've thought about it.

None of this changes the shippabilty of this code but perhaps we want to push only the "final" version if we're convinced confidence belongs in deck_cache?

(Also, does archetype_id belong in deck_cache??? I should probably close this can of worms now.)

maintenance/guess_league_archetypes.py
```diff
@@ -3,8 +3,10 @@
 HOURLY = True
 
 def run() -> None:
+    archetype.clear_old_predictions()
     decks = deck.load_decks('archetype_id is null')
     deck.calculate_similar_decks(decks)
     for d in decks:
-        if d.similar_decks:
-            archetype.assign(d.id, d.similar_decks[0].archetype_id, False)
+        if d.similar_decks and d.similar_decks[0].archetype_id is not None:
```
bakert: Should we spider down until we find a similar deck that *does* have an archetype rather than just punting if top match doesn't have one? Top match could be another recent deck that hasn't been categorized and they could both be copies of a categorized deck. If we do this we'd probably want a minimum threshold but I think that's covered by calculate_similar_decks.

maintenance/guess_league_archetypes.py
```diff
@@ -3,8 +3,10 @@
 HOURLY = True
```
bakert, line 3: This code/decision is not part of this PR but shouldn't we be predicting an archetype at INSERT time like we do with Gatherling decks? There's some benefit to not doing it on user time but I'm ok with writes being slow as they are very rare from a user's perspective. We could even fire another process and send the response leaving the other process to complete the prediction work (and we don't care about giving user feedback if it fails). Fine to leave this out of scope :D

bakert (commented):
I had some thoughts. I'm completely fine with shipping this as-is but I'll give you a minute to decide if you want to make the suggested changes.

decksite/data/archetype.py
adambiltcliffe: done

decksite/data/deck.py
adambiltcliffe: I thought we had some of these already, but I'll replace it with a `.format` for now and we can worry about where to draw the line another time :)

maintenance/guess_league_archetypes.py
adambiltcliffe: Done. Since we're showing a confidence score, no real harm in putting in the best we can find and if the confidence score ends up being 10% or whatever then you can ignore the predicted archetype.

maintenance/guess_league_archetypes.py
adambiltcliffe: Opened #5676 as a separate issue.

bakert (commented):
Nice!

PR #5736, opened 2018-11-04
(Start of) rule system for categorising decks

adambiltcliffe:
This PR doesn't add any user-visible functionality to decksite, but it gives us a sandbox to play with to work out what features we need from a rule-based deck classification system.

It creates an additional page at /admin/rules visible to admins and demimods that allows you to create a rule for a given archetype and then edit the existing rules. A rule consists of a list of cards which must be included and a list of cards which cannot be included. Multiple rules may exist for the same archetype, e.g. in the case of two different builds or cards with equivalent functionality like Wildfire and Burning of Xinye. This could be more sophisticated, but it seemed a simpler place to start than trying to build a system of arbitrary boolean expressions.

The page then reports on:

- How many existing decks are covered by a rule
- How many decks have a conflict (two or more rules apply to them)
- How many decks have a rule-assigned archetype which is different to their tagged archetype

By playing with this we should be able to work out how plausible it is to get the first number as high as possible and the second and third down to zero.

Currently you can't delete rules or change the archetype they're assigned to, although that should be easy to add. I'm not sure what additional features we would want in this system, although something to do with colour seems likely - classifying a deck as Sultai Control has an obvious first criterion to check that we can't currently do.


bakert:
Sweeeeeeet!

decksite/data/deck.py
```diff
@@ -615,6 +615,11 @@ def nwdl_join() -> str:
             ) AS dsum ON d.id = dsum.id
     """
 
+def num_decks() -> int:
+    sql = 'SELECT COUNT(DISTINCT id) AS c FROM deck'
```
bakert: I expect MariaDB optimizes it away but the DISTINCT is redundant here.

decksite/data/deck.py
```diff
@@ -615,6 +615,11 @@ def nwdl_join() -> str:
...
+    r = Container(db().select(sql)[0])
+    return r.c
```
bakert: You can replace these two lines with:

    return db().value(sql)

which gets the value of the only column of the only row of your resultset.

decksite/data/rule.py
```diff
@@ -0,0 +1,189 @@
...
+
+from decksite.data import deck
+from decksite.database import db
+from magic.models import Deck
+from shared.container import Container
+
+
+def num_classified_decks() -> int:
+    sql = 'SELECT COUNT(DISTINCT(deck_id)) AS c FROM ({apply_rules_query}) AS applied_rules'.format(apply_rules_query=apply_rules_query())
+    return (db().select(sql))[0]['c']
```
bakert: Same comments here about `DISTINCT` and about `db().value(sql)`

decksite/data/rule.py
```diff
@@ -0,0 +1,189 @@
...
+
+def mistagged_decks() -> List[Deck]:
+    sql = """SELECT
```
bakert: nit. Standard is to start a new line after triple quotes.

decksite/data/rule.py
```diff
@@ -0,0 +1,189 @@
...
+                    deck_id, rule_archetype.name AS rule_archetype_name, tagged_archetype.name AS tagged_archetype_name
+                FROM
+                    ({apply_rules_query}) AS applied_rules
+                JOIN
```
bakert: We always explicitly say `INNER JOIN` or `LEFT JOIN`. Probably because I'm not actually sure what the default is. `INNER`? I'm not sure if that should be a standard practice or not. Would help me :P

decksite/data/rule.py
```diff
@@ -0,0 +1,189 @@
...
+                    archetype AS rule_archetype
+                ON
+                    rule_archetype.id = applied_rules.archetype_id
+                JOIN
+                    archetype AS tagged_archetype
+                ON
+                    tagged_archetype.id = deck.archetype_id
+                WHERE
+                    rule_archetype.id != tagged_archetype.id""".format(apply_rules_query=apply_rules_query())
```
bakert: nit. Standard is to put the closing triple quotes on the line below here. I don't know if we should have standards that are not lint rules but I also don't like writing lint rules either :D

decksite/data/rule.py
```diff
@@ -0,0 +1,189 @@
...
+    deck_ids: List[str] = []
+    rule_archetypes = {}
+    for r in (Container(row) for row in db().select(sql)):
+        deck_ids.append(str(r.deck_id))
+        rule_archetypes[r.deck_id] = r.rule_archetype_name
+    if len(deck_ids) == 0:
+        return []
+    result = deck.load_decks(where='d.id IN ('+','.join(deck_ids)+')')
```
bakert: Would it be insane to try and crowbar this into load_decks instead of double-querying? Probably.

decksite/data/rule.py
```diff
@@ -0,0 +1,189 @@
...
+                JOIN
+                    archetype AS tagged_archetype
+                ON
+                    tagged_archetype.id = deck.archetype_id
+                WHERE
+                    rule_archetype.id != tagged_archetype.id""".format(apply_rules_query=apply_rules_query())
+    deck_ids: List[str] = []
+    rule_archetypes = {}
+    for r in (Container(row) for row in db().select(sql)):
+        deck_ids.append(str(r.deck_id))
```
bakert, line 39: Why do you want the deck_id as a `str`?

decksite/data/rule.py
bakert: nit.

Standard is spaces around binary operators (so "' + ','.join(deck_ids) + ')') but prefer:

```
ids_list = ', '.join(deck_ids)
result = deck.load_decks(where=f'd.id IN ({ids_list})'
```

or (probably worse):

```
result = deck.load_decks(where='d.id IN ({ids_list})'.format(ids_list=', '.join(deck_ids))
```

decksite/data/rule.py
```diff
@@ -0,0 +1,189 @@
...
+        rule_archetypes[r.deck_id] = r.rule_archetype_name
+    if len(deck_ids) == 0:
+        return []
+    result = deck.load_decks(where='d.id IN ('+','.join(deck_ids)+')')
+    for d in result:
+        d.rule_archetype_name = rule_archetypes[d.id]
+    return result
+
+def doubled_decks() -> List[Deck]:
+    sql = """SELECT
```
bakert: Newline after triple quotes.

Indentation goes a little wonky after this line.

decksite/data/rule.py
```diff
@@ -0,0 +1,189 @@
...
+                FROM
+                    ({apply_rules_query}) AS applied_rules
+                JOIN
+                    archetype
+                ON
+                    applied_rules.archetype_id = archetype.id
+                GROUP BY
+                    deck_id
+                HAVING
+                    COUNT(DISTINCT archetype.id) > 1""".format(apply_rules_query=apply_rules_query())
```
bakert: This feels very similar to the mistagged decks query but I don't know if it's worth trying to generalize.

decksite/data/rule.py
```diff
@@ -0,0 +1,189 @@
...
+    if len(deck_ids) == 0:
+        return []
+    result = deck.load_decks(where='d.id IN ('+','.join(deck_ids)+')')
+    for d in result:
+        d.rule_archetype_name = rule_archetypes[d.id]
+    return result
+
+def doubled_decks() -> List[Deck]:
+    sql = """SELECT
+                    deck_id, GROUP_CONCAT(archetype.name) AS concat_archetypes
```
bakert: Strictly speaking in the super spaced out SQL style each column selected gets its own line.

decksite/data/rule.py
```diff
@@ -0,0 +1,189 @@
...
+        return []
+    result = deck.load_decks(where='d.id IN ('+','.join(deck_ids)+')')
+    for d in result:
+        d.concat_archetypes = concat_archetypes[d.id]
+    return result
+
+def load_all_rules() -> List[Container]:
+    result = []
+    result_by_id = {}
+    sql = 'SELECT rule.id as id, rule.archetype_id, archetype.name as archetype_name FROM rule JOIN archetype on rule.archetype_id = archetype.id'
```
bakert: nittiest of nits. Capitalize `AS` keyword.

decksite/data/rule.py
```diff
@@ -0,0 +1,189 @@
...
+                HAVING
+                    COUNT(DISTINCT archetype.id) > 1""".format(apply_rules_query=apply_rules_query())
+    deck_ids: List[str] = []
+    concat_archetypes = {}
+    for r in (Container(row) for row in db().select(sql)):
+        deck_ids.append(str(r.deck_id))
+        concat_archetypes[r.deck_id] = r.concat_archetypes
+    if len(deck_ids) == 0:
+        return []
+    result = deck.load_decks(where='d.id IN ('+','.join(deck_ids)+')')
```
bakert: This stuff here is repeated code for loading decks by a list of ids. Maybe make a helper function on deck that does that and avoid the duplication?

decksite/data/rule.py
```diff
@@ -0,0 +1,189 @@
...
+        sql = 'INSERT INTO rule_card (rule_id, card, include) VALUES (%s, %s, TRUE)'
+        db().execute(sql, [rule_id, card])
+    for card in exc:
+        sql = 'INSERT INTO rule_card (rule_id, card, include) VALUES (%s, %s, FALSE)'
+        db().execute(sql, [rule_id, card])
+    db().commit('update_rule_cards')
+
+# Currently we do this query several times in a row, but at least with a small number of rules it's cheap enough not to matter
+def apply_rules_query(deck_query: str = '1 = 1'):
+    return f"""
```
bakert, line 113: If you don't want leading whitespace on your SQL you can use the normal indentation for where you are in the code but then call `textwrap.dedent` on the resulting string.

decksite/sql/47.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS rule (
```
bakert: `rule` feels pretty generic. `archetype_rule`? Or does that imply it's a join table of archetypes and rules?

decksite/sql/47.sql
```diff
@@ -0,0 +1,13 @@
...
+  id INTEGER PRIMARY KEY AUTO_INCREMENT,
+  archetype_id INTEGER NOT NULL,
+  FOREIGN KEY(archetype_id) REFERENCES archetype(id) ON DELETE CASCADE
+);
+
+CREATE TABLE IF NOT EXISTS rule_card (
+  id INTEGER PRIMARY KEY AUTO_INCREMENT,
+  rule_id INTEGER NOT NULL,
+  card VARCHAR(100) NOT NULL,
```
bakert: We use VARCHAR(190) for card name columns as that the maximum size of a varchar in the utf8mb4 encoding.

Our Market Research Shows That Players Like Really Long Card Names So We Made this Card to Have the Absolute Longest Card Name Ever Elemental is actually 142 chars, not that it's going to be defining any archetypes any time soon.

decksite/templates/editrules.mustache
```diff
@@ -0,0 +1,88 @@
+<h1>Deck Classification Rules</h1>
+<section>
+    <p>{{num_classified}} out of {{num_total}} decks currently covered by a rule.</p>
+
+    {{#has_doubled_decks}}
```
bakert, line 5: nit. Spacing in templates should follow programmatic rules over making the HTML look perfect in "view source". So as this is functionally an if statement the contents should be indented four spaces.

decksite/templates/editrules.mustache
```diff
@@ -0,0 +1,88 @@
...
+                    <td><a href="{{url}}" title="{{decklist}}">{{name}} ({{id}})</a></td>
+                    <td>{{archetype_name}}</td>
+                    <td>{{concat_archetypes}}</td>
+                </tr>
+            {{/doubled_decks}}
+        </tbody>
+    </table>
+    {{/has_doubled_decks}}
+
+    {{#has_mistagged_decks}}
```
bakert, line 29: Same.

decksite/templates/editrules.mustache
```diff
@@ -0,0 +1,88 @@
...
+    <h2>Other reports</h2>
+    <p>Not currently reported: Archetypes with no matching rule, list of decks not covered by any rule.
+    (These could be added if and when the list of rules is closer to being complete.)</p>
+
+</section>
+<section>
+
+    <h2>View and edit rules</h2>
+
+    {{#rules}}
```
bakert, line 62: Same.

bakert (commented):
This is pretty sweet. I've made a lot of cosmetic and mostly unimportant suggestions but super glad to see this taking off :D

decksite/sql/47.sql
adambiltcliffe: Previously you suggested a single-word name for tables that represent a single entity. Of course, arguably a rule is a join of an archetype and a rule, we just don't bother tracking the 'rule' part because it contains zero information and the relationship is one-to-many. I don't know if we're ever sufficiently likely to have anything else called `rule` that this has any potential ambiguity to make it worth making all the queries even more verbose.

silasary:
What do rules look like?

decksite/sql/47.sql
bakert: Yeah, neither thing makes me totally happy. Rule is just such a very general word. But I guess there's not much point in changing it to something longer which is also problematic.

Are any of classification/codification/systematization/codification/allocation/designation/regulation/assignment better? Probably not.

bakert:
I had a whole bunch of second-order thoughts about this. Mostly about "priority" of both rules and cards. And colors. But I'm going to let it all percolate a bit and maybe have a fiddle with what you have built so far. I still think this is going to be super awesome for getting every Cloudy with Sunny Spells or Reliving Ultimatums deck into the right place very-close-to-automatically even if we don't do crazy more-complicated stuff like saying "this looks like a Grixis Control deck" to the first Grixis Control deck of a new season or whatever. It's gonna be GREAT! :D

silasary:
I posted this in #code a while back, but Gatherling's old system used archetype definitions like [this one](https://raw.githubusercontent.com/PennyDreadfulMtg/gatherling/master/util/archetypes/Spore%20Cry.txt)

The sorter finds all archetypes where the deck matches all **Required** cards.
It then scores the remaining archetypes, where each Strong is 5 points, each moderate is 3, and each weak is 1.
The archetype with the highest score is then used.

decksite/data/rule.py
adambiltcliffe: The DISTINCT isn't redundant here if we're in the situation where we have a deck which currently fits multiple rules, right?

decksite/data/rule.py
adambiltcliffe: `INNER` is definitely default, but I don't mind being explicit

decksite/data/rule.py
bakert: Oh, this is more complicated than I thought, then :D

adambiltcliffe: The result set of the SQL contains one row for each `(deck_id, rule_id)` pair that could apply (it also contains the `archetype_id` for convenience, but that information can be extracted from the `rule_id`).

adambiltcliffe:
I'm a bit wary about a scoring/weighting-based system just because there are a lot of parameters to fine-tune and if it's getting the results wrong it's not always clear what the appropriate way to fix it is. We may well need something more fuzzy as this develops but I'd like to start off by seeing how much of the deck-space we can cover with straightforward boolean rules.

decksite/data/rule.py
adambiltcliffe: Let's keep them separate at least until the rules system is more mature - if we keep them less tightly coupled for now, we can back out parts of the rule system more easily if they're not what we want.

decksite/data/rule.py
adambiltcliffe: Sounds sensible but I'll open it as a separate issue rather than sidetrack this

PR #4705, opened 2018-04-27
Enable `mypy --disallow-incomplete-defs`

silasary:
This PR is the next step in getting to static typing.

It's also a long way away.

PR #5771, opened 2018-11-09
100% typing

silasary:


silasary:
Still not actually 100%, but 100% enough to turn on #4705

bakert:
wooooop!

PR #5906, opened 2018-12-21
Tentative steps towards MTGJSON4. This will probably fail.

bakert:


bakert:
I'm killing this as I diverged. There's probably at least one thing in the fixes that we still need but I tried to manually move most of it across when I made my new branch.

PR #5907, opened 2018-12-24
Issue 5880 - Add archetype pages which exclude League results

tsutton:
- The data is stored as new columns in `_archetype_stats` and `_matchup_stats` tables, possibly slightly increasing the time it takes to rebuild cache.
- The URLs I have chose are `/archetypes/tournament/` and `/archetypes/tournament/<id>/`  (and `seasons/<id>/those`).
- The functionality is built into existing functions and views with an additional argument, `tournament_only`, defaulting to `False` to preserve existing code.

Feedback very welcome :).


decksite/data/archetype.py
```diff
@@ -17,7 +17,7 @@ class Archetype(Container, NodeMixin):
 
 BASE_ARCHETYPES: Dict[Archetype, Archetype] = {}
 
-def load_archetype(archetype: Union[int, str], season_id: int = None) -> Archetype:
+def load_archetype(archetype: Union[int, str], season_id: int = None, tournament_only: bool = False) -> Archetype:
```
bakert: What about always pulling out both sets of stats? If they're going to be precalculated anyway it doesn't really cost anything in terms of user time and we can skip this parameter.

decksite/data/archetype.py
```diff
@@ -87,7 +90,8 @@ def load_archetypes_deckless(order_by: str = '`num_decks` DESC, `wins` DESC, nam
             aca.ancestor -- aca.ancestor will be unique per a.id because of integrity constraints enforced elsewhere (each archetype has one ancestor) but we let the database know here.
         ORDER BY
             {order_by}
-    """.format(season_query=query.season_query(season_id), order_by=order_by)
+    """.format(tournament='_nonleague' if tournament_only else '',
```
bakert: Although it's irrelevant given that all our match data is from league or tournaments right now maybe call this `_tournaments` not `_nonleague`?

decksite/data/archetype.py
```diff
@@ -35,8 +35,11 @@ def load_archetype(archetype: Union[int, str], season_id: int = None) -> Archety
         arch.decks = []
     return arch
 
-def load_archetypes(where: str = '1 = 1', merge: bool = False, season_id: int = None) -> List[Archetype]:
-    decks = deck.load_decks(where, season_id=season_id)
+def load_archetypes(where: str = '1 = 1', merge: bool = False, season_id: int = None, tournament_only: bool = False) -> List[Archetype]:
```
bakert: Similarly here I'd just load both and have the caller decide which it wants to use.

decksite/data/archetype.py
```diff
@@ -35,8 +35,11 @@ def load_archetype(archetype: Union[int, str], season_id: int = None) -> Archety
...
+    having = '1 = 1'
+    if tournament_only:
+        having = 'ct.name != "League"'
```
bakert: I'd go with `= 'tournament'` over `!= 'League'` although as discussed the meaning is currently the same.

Nit. Our standard quote in SQL is single quote.

decksite/data/archetype.py
```diff
@@ -35,8 +35,11 @@ def load_archetype(archetype: Union[int, str], season_id: int = None) -> Archety
...
+    decks = deck.load_decks(where, season_id=season_id, having=having)
```
bakert: Why does this need to be in the HAVING rather than in the WHERE? Each deck has exactly one competition_type so I think it's possible to get it into the WHERE clause even though the query uses a GROUP BY?

decksite/main.py
```diff
@@ -144,13 +144,36 @@ def archetypes() -> str:
     view = Archetypes(deckless_archetypes, all_matchups)
     return view.page()
 
+@APP.route('/archetypes/tournament/')
```
bakert, line 147: Alternative approach is to load everything into /archetypes/ and flip between the two with js or something. Given page size concerns and addressibilty I think I like the way you've gone here but could be swayed.

decksite/main.py
```diff
@@ -144,13 +144,36 @@ def archetypes() -> str:
...
 def archetype(archetype_id: str) -> str:
     season_id = get_season_id()
     a = archs.load_archetype(archetype_id.replace('+', ' '), season_id=season_id)
-    view = Archetype(a, archs.load_archetypes_deckless_for(a.id, season_id=season_id), archs.load_matchups(a.id, season_id=season_id), season_id)
+    deckless_archetypes = archs.load_archetypes_deckless_for(a.id, season_id=season_id)
+    matchups = archs.load_matchups(a.id, season_id=season_id)
+    view = Archetype(a, deckless_archetypes, matchups, season_id)
+    return view.page()
+
+@APP.route('/archetypes/tournament/<archetype_id>/')
```
bakert: Or `/archetypes/<archetype_id>/tournament`. I'm not sure if there's a precedent to follow here with anything else.

bakert (commented):
Cool stuff. A couple of comments but nothing earthshattering. We'll need to have the migration to Scryfall (from mtgjson) working before we can land this.

decksite/data/archetype.py
tsutton: It looks like the competition_type is named 'Gatherling' - so `= 'Gatherling'` instead of `!= "League"`?

decksite/main.py
tsutton: Honestly, I think that it would be nicer to switch like that, but I don't know how to do the JS. Maybe I'll revisit that in the future.

decksite/main.py
tsutton: That makes sense.

decksite/data/archetype.py
tsutton: You're right, I'll switch it to WHERE.

decksite/data/archetype.py
tsutton: Will fix.

decksite/data/archetype.py
tsutton: I've pushed a change to do this. The result is the tournament filtering getting pushed down the pipeline to later in the view processing. The effect of making the file data/archetype.py agnostic to this is contained in [this diff](https://github.com/PennyDreadfulMTG/Penny-Dreadful-Tools/pull/5907/commits/451c745ffc36992d1b634263f097b7b18b40d831).

tsutton: Actually, now that I think about it, maybe it would be even better to pass all of the data to the template and check at that level which to display. That would be more in line with eventually having a javascript toggle, too.

bakert: That sounds good if it's not any slower (which I theorize it should not be). I'm going to merge this as soon as I'm done with migrating to scryfall from mtgjson3. Should hopefully not be too long. I'm down to the level of dealing with weird card types like "Scheme" now. Most everything normal is working.

tsutton: The current version passes both sets of info to the template and lets the template decide which to display. The code is a little uglier now, but it's more flexible in theory, for example if a page wants to have both tournament and all results side by side, or if a future version of this will load both and switch via js instead.

bakert:
Pull from master and push and we can merge this.

PR #5916, opened 2019-01-05
Assorted Discordbot bug fixes

silasary:


bakert:
This won't merge because command.py has gone over 1000 lines. Maybe time to break it up? Or ignore the rule.

silasary:
Ignore the rule for now

Issue #5721, opened 2018-11-02
the position numbers on achievement leaderboards no longer correctly account for ties


vorpal-buildbot:
Reported on Discord by acidicrhyme#5579

bakert:
I suspect the prepare_leaderboards code assumes there is always a tie breaker and has no concept of a dead heat. Hopefully not hard to add that.

tsutton:
Fixed by 67f1bda59e3e2b06423a18ab5349c6f337531bb6.

PR #5918, opened 2019-01-05
Fix issue #5721 - Add logic for ties in leaderboards.

tsutton:
This seemed easy enough, so I took a quick stab at it Now, each method that sets a `leaderboard` attribute is now responsible for providing each row with a `finish` attribute in addition to old stuff.

Previously, the `prepare_leaderboard` method was given a pre-sorted listed, assumed there were no ties, and added the `finish` attribute based on order.


decksite/view.py
```diff
@@ -303,13 +303,10 @@ def prepare_leaderboards(self) -> None:
             self.prepare_leaderboard(l)
 
     def prepare_leaderboard(self, leaderboard: List[Container]) -> None:
```
bakert, line 305: Alternative thought. Insist that callers attach a "score" tuple to each entry in `leaderboard` and do the logic once-only in here. Not sure if it's any better.

tsutton: Yeah... that makes sense. Should have realized to do that when I coded the kind of loop which checks for ties 3 times :P. I'll make that change.

bakert:
@tsutton if you merge from master and push to this PR we can merge this now. hurray!

PR #5972, opened 2019-01-25
Revert "Giant hack to workaround whatsinstandard not having War of the Spark."

silasary:
Reverts PennyDreadfulMTG/Penny-Dreadful-Tools#5967


silasary:
For when https://github.com/glacials/whatsinstandard/pull/122 lands.

PR #6002, opened 2019-01-30

silasary:


silasary:
Anytree has a lot of changes, and one of those is causing us *Issues*.

PR #6220, opened 2019-04-12
Scryfall update optimization

tsutton:
This is my attempt to speed up the database update from Scryfall - magic.multiverse.init() takes about 120-125 seconds on my local system. It's not the best possible for sure; for example, it's not incremental at all, it still does the complete database nuke and rebuild. Just, faster than presently. I feel this is significant enough a change to warrant more eyes looking at it and maybe more testing than just the auto tests, so I've added the do not merge tag. Please let me know if you have feedback.

Below are some notes:

- The following tables appear to unused and/or nonfunctional, so references to them have been removed:
    - card_supertypes and card_subtypes - these are unused as far as I can tell, but can be easily restored
    - card_alias appears to be unused and never populated; we load from the text file instead.
- Columns deprecated (no longer populated and probably weren't functioning beforehand)
   - Face: starter; image_name
   - Printing: border; timeshifted; mci_number
- Currently, I have preserved the functionality of storing non-pd format legality (e.g. Standard). This adds a not-insignificant amount of time to the update (15 seconds on local out of 120 for multiverse.init())
- Old PD formats take a significant amount of time (~75 seconds of 120). Not exactly sure what to do about this - since we are wiping the card table, we must also wipe the card_legality table. It could probably be made better somehow, though. As there become more seasons, this'll need to be changed again.


silasary:
Feels good to me.

PR #6276, opened 2019-05-30
Updated "!explain spectating" command

artillect:
No longer says that spectating during sideboarding crashes the client


artillect:
Not sure how this fails because I just changed some plaintext in the python file.

silasary:
This is because https://scryfall.com/card/mh1/224/icehide-golem was released this week, and the database doesn't expect snow in the mana cost

PR #6313, opened 2019-06-25
Adding to /cards/ information about sideboards, and AJAX scryfall search

tsutton:
Also modify the structure of table deck_card to be a little easier to work with


magic/oracle.py
```diff
@@ -35,17 +35,25 @@ def load_cards(names: Iterable[str] = None, where: Optional[str] = None) -> List
...
-        raise TooFewItemsException('Expected `{namelen}` and got `{rslen}` with `{names}`.  missing=`{missing}`'.format(namelen=len(setnames), rslen=len(rs), names=setnames, missing=missing))
-    return [Card(r) for r in rs]
+    rows = db().select(sql)
+    if setnames and len(setnames) != len(rows):
+        missing = setnames.symmetric_difference([row['name'] for row in rows])
+        raise TooFewItemsException('Expected `{namelen}` and got `{rowslen}` with `{names}`.  missing=`{missing}`'.format(namelen=len(setnames), rowslen=len(rows), names=setnames, missing=missing))
+    ret = []
+    for row in rows:
+        c = Card(row)
+        if c.colors:
```
bakert, line 51: This property does not exist so crashes here.

bakert: Or maybe you brought it into existence but we crash before the db migration runs?

tsutton: Okay yeah. I made some changes, then decided to do things a different way and undid them. But I failed to undo them everywhere. I can fix this.

tsutton: Is there a way to apply migrations to the cards database? the magic/ directory doesn't have a sql/ subdir.

bakert: Cards db is entirely dropped and recreated when it changes. You can make the changes to database/multiverse/card and then bump `SCHEMA_VERSION = 106` in database.py if you want to change the schema

PR #6334, opened 2019-06-27
Improvements to /cards/, attempt 3

tsutton:
### **THIS PR IS BEING REWORKED**
This PR is being split into two and redone, do not merge this. I am leaving this open as a reference until the other two are done.

------------------------

The only code differences between this and last time are that I changed `decksite/sql/52.sql` to no longer have a `DROP TABLE`. To be honest, I'm not entirely sure what failed about this branch last time that caused the migration to only partially work.

So I suggest that someone review this by backing up their local DB copy, then checking this out and applying the migration. If that works, then back up the decksite DB on prod and try merging it

The DB structural changes are to `deck_card`, making it so instead of the boolean column `sideboard` to tell whether the row is representing cards in the main or side, each row just has two values, `n_main, n_side`.

Also, the `_cache_card` table gains an additional two columns, `num_deck_sideboard` and `num_deck_sideboard_tournament`, thus the cards and person-cards preaggregation need to be reset.


decksite/data/deck.py
```diff
@@ -488,15 +492,18 @@ def load_cards(decks: List[Deck]) -> None:
         return
     decks_by_id = {d.id: d for d in decks}
     sql = """
-        SELECT deck_id, card, n, sideboard FROM deck_card WHERE deck_id IN ({deck_ids})
+        SELECT deck_id, card, n_main, n_side FROM deck_card WHERE deck_id IN ({deck_ids})
```
bakert, line 495: I think you can avoid all the database schema changing with a query like this:

```
SELECT
    SUM(CASE WHEN sideboard THEN 0 ELSE 1 END) AS n_main,
    SUM(CASE WHEN sideboard THEN 1 ELSE 0 END) AS n_side
FROM
    deck_card
WHERE
    deck_id IN ({deck_ids})
```

Does that do everything you need? Are there performance ramifications?

decksite/sql/52.sql
```diff
@@ -0,0 +1,26 @@
+-- Change deck_card so that each row has the number in main and side in that deck
+-- instead of the old way (each row counts for main or side depending on column sideboard)
+
+-- No table references deck_card, so we'll just delete it and make a new one
```
bakert, line 4: This comment is not valid.

decksite/sql/52.sql
```diff
@@ -0,0 +1,26 @@
...
+    FROM deck_card
+    GROUP BY card, deck_id
+    ORDER BY deck_id;
+
+RENAME TABLE deck_card to _old_deck_card;
+
+RENAME TABLE _new_deck_card TO deck_card;
+
+# Reset the preaggregation
+DROP TABLE IF EXISTS _card_stats, _card_person_stats;
```
bakert, line 26: At the end of these changes all the data will be in _old_deck_card and deck_card will be empty. I assume that's not the desired effect.

decksite/templates/cardstable.mustache
```diff
@@ -6,23 +6,33 @@
...
+            <th class="n" data-breakpoints="xs" data-sortInitialOrder="desc">
+                <abbr title="Tournament wins">①</abbr>
+            </th>
+            <th class="n" data-breakpoints="xs" data-sortInitialOrder="desc">
+                <abbr title="Tournament top-8s">⑧</abbr>
+            </th>
+            <th class="n" data-breakpoints="xs" data-sortInitialOrder="desc">
+                <abbr title="League 5-0 runs">5–0s</abbr>
+            </th>
+            <th class="n" data-breakpoints="xs" data-sortInitialOrder="desc">
```
bakert, line 18: We used to display this information back when we displayed "this season" and "all time" alongside one another but dropped it on the thinking that in general you'd rather see something simpler and faster. I can see the argument that it is interesting enough to include, though. Should the values be "main deck" and "sideboard" rather than "# decks" and "# SBs". Also "# SBs" is kind of an ugly construction but might be a necessary evil. I can't think of anything short and better but that again makes me question the premise.

magic/oracle.py
```diff
@@ -36,17 +36,17 @@ def load_cards(names: Iterable[str] = None, where: Optional[str] = None) -> List
     else:
         setnames = set()
     if setnames:
-        names_clause = 'c.name IN ({names})'.format(names=', '.join(sqlescape(name) for name in setnames))
+        names_clause = 'name IN ({names})'.format(names=', '.join(sqlescape(name) for name in setnames))
```
bakert, line 39: Why do we drop the `c` prefix here?

magic/oracle.py
```diff
@@ -36,17 +36,17 @@ def load_cards(names: Iterable[str] = None, where: Optional[str] = None) -> List
...
         names_clause = '(1 = 1)'
     if where is None:
         where = '(1 = 1)'
     sql = multiverse.cached_base_query('({where} AND {names})'.format(where=where, names=names_clause))
-    rs = db().select(sql)
-    if setnames and len(setnames) != len(rs):
-        missing = setnames.symmetric_difference([r['name'] for r in rs])
-        raise TooFewItemsException('Expected `{namelen}` and got `{rslen}` with `{names}`.  missing=`{missing}`'.format(namelen=len(setnames), rslen=len(rs), names=setnames, missing=missing))
-    return [Card(r) for r in rs]
+    rows = db().select(sql)
```
bakert, line 45: The renaming of the variables here feels a little pointless. `rs` is a pretty traditional name for a resultset/recordset. In general, also, we use 'x' and 'xs' for local vars throughout the codebase.

magic/oracle.py
```diff
@@ -36,17 +36,17 @@ def load_cards(names: Iterable[str] = None, where: Optional[str] = None) -> List
...
+    if setnames and len(setnames) != len(rows):
+        missing = setnames.symmetric_difference([row['name'] for row in rows])
+        raise TooFewItemsException('Expected `{namelen}` and got `{rowslen}` with `{names}`.  missing=`{missing}`'.format(namelen=len(setnames), rowslen=len(rows), names=setnames, missing=missing))
```
bakert, line 48: When changing strings in `'{blah}'.format(blah=blah)` style feel free to switch to `f'{blah}'` style which did not exist when the project began but is now our standard. Not important.

shared_web/static/js/pd.js
```diff
@@ -367,6 +368,70 @@ PD.htmlEscape = function (s) {
...
+    if (query == ""){
+        PD.Filter.reset();
+        return;
+    }
+
+    $("#scryfall-filter-submit").attr("disabled", "disabled").text("Loading...");
+    $("#scryfall-filter-reset").attr("disabled", "disabled").text("Loading...");
+    $("#scryfall-filter-form").submit(function () { return false; })
+
+    query = "f:pd and (" + query + ")";
```
bakert, line 395: This will break when viewing /seasons/{not current season}/cards/ – there will be cards that should match the search that do not because they are not legal in the *current* pd season. Scryfall does not offer historical pd season search.

shared_web/static/js/pd.js
```diff
@@ -367,6 +368,70 @@ PD.htmlEscape = function (s) {
     return $("<div>").text(s).html();
 };
 
+PD.Filter = {};
```
bakert, line 371: Why is `Filter` capitalized?

shared_web/static/js/pd.js
```diff
@@ -367,6 +368,70 @@ PD.htmlEscape = function (s) {
...
+    // set up the event handlers for the form
+    $("#scryfall-filter-form").submit(function () {
+        PD.Filter.scryfallFilter($("#scryfall-filter-input").val());
+        return false;
+    });
+    $("#scryfall-filter-reset").click(PD.Filter.reset);
+};
+
+PD.Filter.scryfallFilter = function (query) {
+    if (query == ""){
```
bakert, line 386: Triple equals for javascript.

shared_web/static/js/pd.js
```diff
@@ -367,6 +368,70 @@ PD.htmlEscape = function (s) {
...
+        PD.Filter.reset();
+        return;
+    }
+
+    $("#scryfall-filter-submit").attr("disabled", "disabled").text("Loading...");
```
bakert, line 391: Please use a "real" ellipsis char here: "…" and not three periods.

shared_web/static/js/pd.js
```diff
@@ -367,6 +368,70 @@ PD.htmlEscape = function (s) {
...
+    $("#scryfall-filter-reset").attr("disabled", "disabled").text("Loading...");
```
bakert, line 392: Please use a "real" ellipsis char here: "…" and not three periods.

shared_web/static/js/pd.js
```diff
@@ -367,6 +368,70 @@ PD.htmlEscape = function (s) {
...
+                jqEle.show();
+            }
+        });
+        $("#scryfall-filter-submit").removeAttr("disabled").text("Search");
+        $("#scryfall-filter-reset").removeAttr("disabled").text("Reset");
+        $("#scryfall-filter-form").submit(function () { return false; })
+    }
+
+    function parse_and_continue (blob) {
+        // TODO: error handling
```
bakert, line 414: Prefer GitHub issues to TODOs. This and a couple of other things on this review should be found by the linter, I think. Try `make lint` in the root of the project on your dev machine. Maybe that skips js?

shared_web/static/js/pd.js
```diff
@@ -367,6 +368,70 @@ PD.htmlEscape = function (s) {
...
+
+        for (i=0; i<blob.data.length; i++) {
+            card_names.push(blob["data"][i]["name"])
+        }
+        if (blob["has_more"]) {
+            $.getJSON(blob["next_page"], parse_and_continue)
+        }
```
bakert, line 421: K&R brace style please.

shared_web/static/js/pd.js
```diff
@@ -367,6 +368,70 @@ PD.htmlEscape = function (s) {
...
+    }
+
+    $.getJSON("https://api.scryfall.com/cards/search?q=" + query, parse_and_continue);
+
+};
+
+PD.Filter.reset = function () {
+    $("tr.cardrow").show();
+};
+
```
bakert, line 434: If we're going to expand this kind of filtering to the cards page we should make it consistent with the filtering on /rotation/. Either they both use javascript or neither do.

How is the performance?

shared_web/static/js/pd.js
bakert: For that matter our standard loading spinner is:

```
<div class="loading"><span class="spinner"></span> <span class="text">Loading…</span></div>
```

see spinner.mustache

bakert (changes requested):
I think we can do what you want in the database without schema changes.

I have some concerns about js filtering of the /cards/ but I think it could be a big improvement on reloading the page especially because /cards/ can be super huge. Will need some tweaks though because historical /cards/ pages can't use `f:pd` filter. I wonder if the filter /rotation/ has this problem – did you copy it from there?

decksite/sql/52.sql
bakert: Dropping the preaggregation tables like this will cause the preaggregation to happen on user time which takes several minutes but if we back out the schema changes that's not a concern as we won't be doing this migration any more.

shared_web/static/js/pd.js
tsutton: I think the performance is basically same, it's gated by making the requests to scryfall and parsing all the different pages. This way, the requests are directly between the user and Scryfall, rather than between prod and Scryfall - so maybe faster or slower. The old way, the whole page has to be reloaded, but this way, jQuery has to go through the page modifying the DOM. Also, doing it with JS means there's no extra hits to the database, but those hits aren't that expensive since it's all coming from cached queries.

I personally think doing it in JS is snazzier and I envisioned it being nice for future parts of this page, like e.g. if I can make a search for cards fitting a certain Scryfall fliter (like the colors of my deck) and then do different sub-filters like cards good vs archetype 1, cards good vs archetype 2. But I don't feel strongly. I do agree that /cards/ and /rotation/ should be unified, though.

decksite/data/deck.py
tsutton: The reason I wanted to do this change is because I couldn't figure out how to get `decksite/data/card.py:preaggregate_card()` to work. I think I was having trouble with the GROUPing. But I wil see if I can get it to work.

tsutton:
You are totally right about the `f:pd` and historical formats. I was trying to do that as a cute optimization. It wasn't copied.

shared_web/static/js/pd.js
tsutton: I think for specific queries it can be faster with JS. Something like `t:creature` or `c:u` will be about the same, but something like `t:land` is probably quicker with JS.

shared_web/static/js/pd.js
tsutton: Sorry, yeah, the linter didn't catch that.

decksite/data/deck.py
tsutton: I got it to work now, I don't recall exactly what the thing blocking me before was, but this works fine.

decksite/templates/cardstable.mustache
tsutton: My original motivation for adding this was to help myself and others build their sideboards. The /cards/ page could theoretically help by showing what cards other people are playing in their sideboards. However, I wanted to keep it All Decks and Sideboards (rather than #main and #side) to preserve the chart's sorting as overall cards by popularity. 

Totally open to #SBs being changed to something else, part of the reason I added the `<abbr>` tags is because I didn't like how that looked.

bakert: Yeah I think that's a very real motivation. I have experienced it myself. I'm trying to think if there's a better way to satisfy that need. I can envisage a /cards/sideboard/ but that's pretty weird. Maybe this is the best way. Worth discussing in #code or even #general to see if anyone has a smart idea? An additional column is not such a huge thing especially with gzip but it's worth canvassing a few ideas maybe.

bakert: I wonder how bad "# sideboards" looks. It's very long compared to most of the headings. "# side"??

shared_web/static/js/pd.js
bakert: I'm not against a js version. It seems like it will probably be a significantly better experience as long as we handle super slow queries/slow connections well. I'm not sure how I feel about querying the scryfall API directly and not via `fetcher`. I guess it's ok/better than going an additional hop through some endpoint on pdm. Not sure about that. How hard to swap this in on /rotation/ too?

PR #6341, opened 2019-07-01
Async filtering to /cards/ and /rotation/ based on Scryfall Queries

tsutton:
This pull request is built from parts of #6334 and improvements since then. It adds a form to /cards/ and /rotation/ (including /rotation/new/ etc) which lets the user input a query, then submits that to Scryfall and filters the page in-place. It supports browser history as well.


decksite/templates/cards.mustache
```diff
@@ -1,4 +1,8 @@
 {{> banner}}
 <section>
+    <button id="toggle-filters-button" onclick="PD.filter.toggleDisplayFilter()">Show filters</button>
```
bakert: Please attach the onclick behavior in pd.js don’t inline it here. It’s our standard.

decksite/templates/cardstable.mustache
```diff
@@ -6,23 +6,33 @@
             <th class="n" data-sortInitialOrder="desc"># Decks</th>
             <th class="n" data-breakpoints="xs" data-sortInitialOrder="desc">Record</th>
             <th class="n" data-breakpoints="xs" data-sortInitialOrder="desc">Win %</th>
-            <th class="n" data-breakpoints="xs" data-sortInitialOrder="desc">①</th>
-            <th class="n" data-breakpoints="xs" data-sortInitialOrder="desc">⑧</th>
-            <th class="n" data-breakpoints="xs" data-sortInitialOrder="desc">5–0s</th>
+            <th class="n" data-breakpoints="xs" data-sortInitialOrder="desc">
+                <abbr title="Tournament wins">①</abbr>
```
bakert, line 10: If we’re going to do this we should do it everywhere that we use these symbols.

decksite/main.py
```diff
@@ -320,9 +323,10 @@ def community_guidelines() -> str:
 @APP.route('/rotation/<interestingness>/')
 @cached()
 def rotation(interestingness: Optional[str] = None) -> str:
-    rotation_query = request.args.get('rq')
-    _, cardnames = fetcher.search_scryfall(rotation_query, exhaustive=True) if rotation_query else (None, None)
-    view = Rotation(interestingness, rotation_query, cardnames)
+    query = request.args.get('fq')
+    if query is None:
+        query = ''
```
bakert, line 328: If it’s easy it would be good to add this to /rotation/changes and /rotation/speculation too. That can definitely be in a follow up PR tho.

decksite/templates/cardstable.mustache
```diff
@@ -6,23 +6,33 @@
...
+                <abbr title="Number of decks with card in the sideboard"> # SBs</abbr>
+            </th>
             {{! Empty column to hold the toggle when footable kicks in.}}
             <td></td>
         </tr>
     </thead>
     <tbody>
         {{#cards}}
-            <tr>
+            <tr class="cardrow" id="cardrow-{{name}}">
```
bakert: I’m not a big fan of ids in html in general. Things you expect to remain unique have a tendency to become ununique. In this case it’s pretty easy to imagine a page with two cardtables on it where the same card appears in both tables.  Can we make this a class?  I’d also sanitize the name down to lowercase letters and hyphens only personally but that’s less important. Hopefully gzip means this doesn’t really add much page weight to something like /seasons/all/cards but might be worth a spot check.

decksite/templates/rotation.mustache
```diff
@@ -4,14 +4,12 @@
...
-        {{#rotation_query}}
-            <p>
-                <form class="inline">
-                    <input type="text" name="rq" value="{{rotation_query}}">
-                    <button type="submit">Search</button>
-                </form>
-            </p>
-        {{/rotation_query}}
+        {{#in_rotation}}
+            <button id="toggle-filters-button" onclick="PD.filter.toggleDisplayFilter()">Show filters</button>
```
bakert: If this is identical each time it shows up let’s make it a partial.  You can include partials

{{> likethis}}

shared_web/static/js/pd.js
```diff
@@ -367,6 +368,164 @@ PD.htmlEscape = function (s) {
...
+        let jqEle = $(this);
+        if (cardNames.indexOf(this.id.split('-')[1]) == -1){
+            jqEle.hide();
+        } else {
+            jqEle.show();
+        }
+    });
+};
+
+// input url returns a promise to {success: true/false, cardNames: [...], error message: {...}
```
bakert: Nit. You are missing a closing curly brace on the end of this comment.

decksite/views/cards.py
```diff
@@ -9,11 +9,13 @@
...
 # pylint: disable=no-self-use
 class Cards(View):
-    def __init__(self, cards: List[Card], tournament_only: bool = False) -> None:
+    def __init__(self, cards: List[Card], tournament_only: bool = False, query: str = '') -> None:
         super().__init__()
         self.show_seasons = True
         self.show_tournament_toggle = True
         self.tournament_only = tournament_only
+        self.query = query
+        self.cards = cards
```
bakert: It’s surprising to me that this view wasn’t setting self.cards already. Without that how were the cards getting vivified for display? How were they being shown in the mustache. Not sure what problems this might cause but it made my spidey sense tingle.

decksite/templates/cardstable.mustache
```diff
@@ -6,23 +6,33 @@
...
+                <abbr title="Tournament wins">①</abbr>
+            </th>
+            <th class="n" data-breakpoints="xs" data-sortInitialOrder="desc">
+                <abbr title="Tournament top-8s">⑧</abbr>
+            </th>
+            <th class="n" data-breakpoints="xs" data-sortInitialOrder="desc">
+                <abbr title="League 5-0 runs">5–0s</abbr>
+            </th>
+            <th class="n" data-breakpoints="xs" data-sortInitialOrder="desc">
+                <abbr title="Number of decks with card in the sideboard"> # SBs</abbr>
```
bakert: Is this value already being set or is that in another PR?

bakert:
Cool stuff. I look forward to seeing it in action. I’m still not completely sold on “# SBS” (neither the text nor the existence of the column) but let’s get it out there. 

My comment on ids should also be applied to the query form. I can easily see that being included at both the top and bottom of a big list of cards.

decksite/views/cards.py
tsutton: This line was unnecessary, self.cards was being set a few lines lower already.

decksite/templates/cardstable.mustache
tsutton: Changed to not using id. Everything is stuffed into classes, except I put the card name into a data attribute, which seemed like a fancy thing that I just learned about.

tsutton: The size of the html doc changes by about 10%, and pd.js is about 30% bigger: https://imgur.com/a/7LN6WyY

decksite/main.py
tsutton: Good idea, it shoudn't be hard. I'll make it in a follow up.

tsutton:
@bakert Does this look good now? If so, let's merge it?

PR #6350, opened 2019-07-02
Update the database when scryfall has new bulk data in a less disruptive way.

bakert:


bakert:
I'm adding "do not merge" label here because this makes me nervous. Local testing says it's fine but want someone to say "I can't see anything obviously terrible here" before I merge :D @silasary 
?

bakert:
Crap, this has all the changes from my other PR in it. I'll kill this and make one with just the change I want reviewed. Hurry up jenkins!

PR #6351, opened 2019-07-02
Avoid dropping _cache_card until after a database update is completed.

### LINKED ISSUE #6345: We get little bursts (daily, all at the same time) of 500 errors because "table _cache_card does not exist". This should not happen

bakert:
This will prevent the site from going down for a couple of minutes during
a scryfall update (which are daily when spoilers for new sets are being
releaesed).

Fixes #6345.

### DIFF
--- a/magic/multiverse.py
+++ b/magic/multiverse.py
@@ -127,8 +127,7 @@ def base_query_lite() -> str:
 def update_database(new_date: datetime.datetime) -> None:
     db().begin('update_database')
     db().execute('DELETE FROM scryfall_version')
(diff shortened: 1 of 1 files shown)


bakert:
I'm adding "do not merge" label here because this makes me nervous. Local testing says it's fine but want someone to say "I can't see anything obviously terrible here" before I merge :D @silasary 
?

tsutton:
I tested this a bit locally, including having the update code pause at some key points and making sure the website still worked, and it all looks good to me.

tsutton (approved):

PR #6359, opened 2019-07-03
Add regular rotation messages from the bot during rotation season.

bakert:
This meant pushing a lot of what was happening in decksite.views.rotation
into magic.rotation.

Unfortunately it is not possible to push interestingness into magic.rotation
because it needs knowledge only decksite has.

So the view still decorates the cards with interestingness after the
magic.rotation stuff.

And the bot doesn't know how interesting cards are and just lists an arbitrary
four.

It would be better to expose interestingness to the bot somehow but that will
be a future improvement.

### DIFF
--- a/decksite/view.py
+++ b/decksite/view.py
@@ -119,7 +119,7 @@ class View(BaseView):
         return inflect.engine().number_to_words(len(tournaments.all_series_info()))
 
     def rotation_text(self) -> str:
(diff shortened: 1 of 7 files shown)


bakert:
A "kill switch" for this feature is available by setting rotation_hype_channel_id to `null` or any invalid value in config.json.

PR #6421, opened 2019-07-17
Introduces the p1p1 command.

Volvary:
-Assembles a Pack 1, Pick 1 game for members to discuss.
-Includes a protection to stop multiple p1p1 from happening at the same
time.

Notes: This is a first draft, from what I could cobble in one night from the code. If I can improve something, please give me a reference function I can look at to improve this.
Also, I'm interested to know if a minimum delay between 2 p1p1 commands should be implemented (I figured that we could trust the PD community but I'm open to adding it before implementation.)


Volvary:
image

The bot can take a moment to assemble the query but it produces a nice and fast to appear result.

silasary:
Yeah, looks good to me.

PR #6490, opened 2019-08-06
Prevent decimal point on "hits needed" during rotation.

bakert:


silasary:
Uhhhh? There's some dissonence between title and contents here

bakert:
Changed the title. Must have had a leftover .PULL_REQUEST file (or whatever it is the commandline github tool) from a previous PR  uses when making a pull request.

PR #6533, opened 2019-09-11
Twice in one week, I'm removing this hardcoded list of layouts

silasary:


magic/multiverse.py
```diff
@@ -225,12 +225,10 @@ def insert_cards(printings: List[CardDescription]) -> None:
         cards[p['name']] = card_id
         card_values.append("({i},'{l}')".format(i=card_id, l=p['layout']))
 
-        if p['layout'] in ['augment', 'emblem', 'host', 'leveler', 'meld', 'normal', 'planar', 'saga', 'scheme', 'token', 'vanguard']:
-            face_values.append(single_face_value(p, card_id))
-        elif p['layout'] in ['adventure', 'double_faced_token', 'flip', 'split', 'transform']:
+        if p.get('card_faces'):
```
bakert, line 228: This seems like a reasonable way to make this decision but note that there's a hardcoded list of layouts in `multiverse.layouts` which determines whether we do or do not care about a particular layout. This would probably be a more appropriate place to decide we don't care about art_series rather than the hardcoded exclusion in a later commit.

bakert: #6536

PR #6601, opened 2019-10-24
Five 'nothing has changed' messages in a row makes me sad

silasary:


bakert:
fwiw I prefer the previous behavior. It's once an hour. If it's several messages in a row then it's just the case that no one was chatting. This feels like it's "not working".

bakert:
Not a big deal.

PR #6665, opened 2019-11-02
Switch /decks/ to use a react component for speed/usability.

bakert:


silasary:
Got some nasty scrollbars on Windows/Firefox
image

PR #6717, opened 2019-11-06
Automatic client CDN dependency updates, maybe

bakert:


bakert:
@silasary does what I'm doing here with the subprocess git commands look sane? It works on local but I really have no idea what it's going to do on prod ;)

maintenance/client_dependencies.py
```diff
@@ -0,0 +1,61 @@
...
+PATH = 'shared_web/templates/jsdependencies.mustache'
+
+def run() -> None:
+    tags = [fetch_script_tag(library) + '\n' for library in get_dependencies()]
+    output = ''.join(tags)
+    write_dependencies(output)
+    subprocess.call(['git', 'add', PATH])
+    if subprocess.call(['git', 'commit', '-m', '"Update client dependencies."']) == 0:
+        subprocess.call(['git', 'push'])
+        subprocess.call(['hub', 'pull-request', '-b', 'master', '-m', 'Update client dependencies.' '-f'])
```
bakert: This is the part I'm really unsure of.

silasary: This should be an ad-hoc, and should never run on prod.
I can wrap it into the jenkins stuff

bakert: Ah, yeah, that makes so much more sense.

bakert: 0292c632 does this.

Issue #5471, opened 2018-10-03
/rotation is slow because of read_files and prepare_card


bakert:

tsutton:
`read_files` is mostly handled by #6688 . As far as `prepare_card`, I feel like the crux of the problem is that the function is super overloading, preparing cards for what any card might need on any page. Maybe it should be split up with a view being able to configure which parts of cards need preparing? As an extremely barebones proof of concept, if I comment out various parts of `prepare_card` I can get it down by quite a bit for `/rotation`, but there might be some overhead if there's various checks against which components are needed.

The other idea would be to try and batch the card preparation, i.e. right now we call `url_for` for each card, whereas we could call it once for all cards to get a format string and plug the cardname into that.

The other thing is, it takes a long time for pystache to render all the cards. Locally, on a hard refresh, it is the dominant component by far. This may not be a realistic representation of prod, though, f it would be helped by the `CachedRenderer` magic in https://github.com/PennyDreadfulMTG/Penny-Dreadful-Tools/blob/master/shared_web/template.py if I weren't doing a hard refresh. Apparently `pystache` is basically abandoned, and https://github.com/noahmorrison/chevron is supposed to be faster, but chevron is much more basic, and it might be hard to migrate over.

tsutton:
For reference, here's what I'm looking at: three runs of `/rotation/` with 11012 cards each in 50 identical `Run_*.txt` files. Redis fully primed, code set to after #6688 .
```
** 1 - 4.98s
PATH: '/rotation/'
         6459734 function calls (5150999 primitive calls) in 4.976 seconds

   Ordered by: internal time, call count
   List reduced from 786 to 30 due to restriction <30>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   418456    0.403    0.000    0.403    0.000 /usr/lib/python3.7/site-packages/munch/__init__.py:109(__setattr__)
        3    0.374    0.125    0.374    0.125 /usr/lib/python3.7/json/decoder.py:343(raw_decode)
    11012    0.197    0.000    1.343    0.000 /home/taylors/projects/Penny-Dreadful-Tools/decksite/view.py:221(prepare_card)
  50760/1    0.144    0.000    3.646    3.646 /usr/lib/python3.7/site-packages/pystache/parsed.py:37(render)
[...]

PR #6773, opened 2019-11-09
Slightly sketchy speed-up for /rotation/

tsutton:
Context: #5471. 

Changes: call Flask's builtin `url_for` once to retrieve a template url and then just use string formatting to build the url for each card.


bakert (commented):
What's the total speedup late in full rotation week? About 1s?

What do you think about moving this to shared_web.base_view.BaseView.url_for or (probably more appropriate) decksite.view.View.url_for and doing a little bit of magic based on the input params. Then we could change literally every url_for to use the new function and do this for anything we see more than a few dozen times in a request?

bakert:
We could even have it store counts of (template + '|' + len(args)) occurrences and flag anything over 50 or something.

PR #6796, opened 2019-11-12
Enforce mypy strict everywhere and make it the only option in dev.py.

bakert:


silasary:
Don't actually depreciate strict, just move the additional options there into the main, and add a new one to strict

silasary:
The next on my list is `--warn-return-any`

bakert:
Landing failure here is because

https://api.scryfall.com/sets 

seems to think that mystery booster playtest cards belong to a set called cmb1

https://scryfall.com/sets/cmb1

but 

https://archive.scryfall.com/json/scryfall-default-cards.json

thinks that the set is called mystpt

for example Taiga Stadium

I have pinged the scryfall folks

bakert:
Bulk data will be updated with correct code tomorrow.

PR #6862, opened 2019-11-18
Downloadable MTGO-friendly files with 4 of each rotated in/out card.

bakert:


bakert:
I think I closed this instead of the ticket.

PR #6881, opened 2019-11-24
Ugly hack while Scryfall set data is buggy.

bakert:


bakert:
This is BS. I am struggling with the new rebase future. Killing this and will make a better PR.

PR #6884, opened 2019-11-24
First proper stake in the ground for game log analysis.

bakert:


dev.py
```diff
@@ -139,10 +139,10 @@ def mypy(argv: List[str], strict: bool = False) -> None:
     if result[2]:
         raise TestFailedException(result[2])
 
-def unit(argv: List[str]) -> None:
-    test(argv, 'not functional and not perf')
+def unit(argv: List[str]):
+    pytest(argv, 'not functional and not perf')
 
-def test(argv: List[str], m: str) -> None:
+def pytest(argv: List[str], m: str) -> None:
```
silasary: We import pytest as a dependency,  this is a naming conflict.

PR #7053, opened 2020-02-13
even more rotation automation

silasary:


magic/fetcher.py
```diff
@@ -124,15 +124,15 @@ def legal_cards(force: bool = False, season: str = None) -> List[str]:
...
+    encoding = 'utf-8'
     cached_path = os.path.join(os.path.abspath(os.path.dirname(__file__)), 'legal_cards')
     if os.path.exists(os.path.join(cached_path, url)):
         h = open(os.path.join(cached_path, url), encoding=encoding)
         legal = h.readlines()
         h.close()
         return [l.strip() for l in legal]
 
-    url = 'http://pdmtgo.com/' + url
+    url = 'https://pennydreadfulmtg.github.io/' + url
```
bakert, line 135: What's the benefit of using the github address? You're staging there?

silasary: Yeah, and the rotation script automatically pushes there, whereas S3 needs manual intervention.

It means our internal tools will be faster and require less human intervention.

PR #7126, opened 2020-03-28
Fix aliases for Gatherling<->MTGO (for admins).

### LINKED ISSUE #7125: Adding an alias of Smell_Hole to (mtgo) Smell-Hole does not work through /admin/aliases but gives no error

bakert:
Add missing param to fill_form annotation that was stopping aliases working.

Also rearrange the order of annotations - security MUST come last to work.

Fixes #7125.


silasary:
> security MUST come last to work.

Nah, it just need to be executed inside of `APP.route`

PR #7186, opened 2020-04-05
og:image meta tag on every page for reddit thumbnails, etc.

bakert:


bakert:
Checking /seasons/ in the smoke test triggers a bunch of the preaggregation which fails to get past the stricter is_full_group_by in Travis CI. I'm not sure why Travis db settings differ from prod.

bakert:
Simplest fix here is to just remove /seasons/ from smoke_test but I'll pursue this elsewhere so we can re-add it.

PR #7228, opened 2020-04-16
Let demimods see the admin menu.

bakert:


decksite/sql/58.sql
```diff
@@ -0,0 +1 @@
+CREATE UNIQUE INDEX idx_u_name ON archetype (name(142));
```
silasary, line 1: Failed on Prod DB

dev.py
```diff
@@ -252,13 +252,13 @@ def build() -> None:
         pipargs.append('--user')
     subprocess.check_call(pipargs)
     print('>>>> Installing node modules')
-    subprocess.check_call(['npm', 'install'], shell=True)
```
silasary, line 255: This breaks windows, as npm does not work without shell=True.
I guess we need a platform check.

PR #7265, opened 2020-04-24
I got bored and fixed things

silasary:


bakert:
Price grabber now says

```
Error: No such command 'price_grabber'.
```

PR #7298, opened 2020-04-30
Correct invalid label in rollback.

bakert:


bakert:
Travis has approved this but github doesn't know for some reason.

PR #7417, opened 2020-06-07
Fix devenv issues

silasary:


.eslintrc.js
```diff
@@ -75,10 +75,7 @@ module.exports = {
         "key-spacing": "off",
         "keyword-spacing": "off",
         "line-comment-position": "off",
-        "linebreak-style": [
-            "error",
-            "unix"
-        ],
+        "linebreak-style": "off",
```
bakert, line 78: Scary. What do we need this for?

dev.py
```diff
@@ -114,14 +115,20 @@ def lint(argv: List[str]) -> None:
...
-            '-j', '4' # Use four cores for speed.
+            '-j', str(configuration.get_int('pylint_threads')), # Use four cores for speed.
            ]
     args.extend(argv or find_files(file_extension='py'))
     # pylint: disable=import-outside-toplevel
     import pylint.lint
-    linter = pylint.lint.Run(args, exit=False)
+    try:
+        linter = pylint.lint.Run(args, exit=False)
+    except PicklingError:
```
bakert, line 126: Oh man I did not consider this as a path forward. Nice!

.eslintrc.js
silasary: Because when git.autocrlf is (correctly) turned on for a Windows machine, it fails the linter.

PR #7551, opened 2020-07-25
Hide intro box after 10 page views.

bakert:


bakert:
This PR is kind of broken. Most of the stuff on it is already in master. Closing it to recreate.

PR #7632, opened 2020-08-13
Add bot command explain netdecking

### LINKED ISSUE #7629: !explain netdecking to codify that borrowing someone's list is a good thing, and that we as a community support doing so
Reported on Discord by silasary#0001

kzcheng:
Fix #7629

### DIFF
--- a/discordbot/commands/explain.py
+++ b/discordbot/commands/explain.py
@@ -71,6 +71,15 @@ async def explain(ctx: MtgContext, *, thing: Optional[str]) -> None:
                 'Current League': fetcher.decksite_url('/league/current/')
             }
         ),
+        'netdecking': (
+            """
+            If you are having trouble coming up with an original decklist, feel free to start by learning from the best!
+            You can find the best tournament winning decks in the link below. Sort by stars to get decks that won tournaments or did well in tournaments and leagues.
+            """,
+            {
+                'Decks': fetcher.decksite_url('/decks/'),
+            }
+        ),
         'noshow': (
             """
             If your opponent does not join your game please @-message them on Discord and contact them on Magic Online.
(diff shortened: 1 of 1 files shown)


bakert (changes requested):
The motivation for this is to be used:

(a) When someone says they feel bad netdecking but they aren't sure they can build something competitive.

or:

(b) When someone says netdecking is evil or some other indefensible attitude from the past.

(a) is much more common than (b) although we have seen both.

I like something short and sweet.

> Netdecking is not only allowed, it is encouraged! Deck creators are flattered not annoyed when others play their decks!

discordbot/commands/explain.py
```diff
@@ -95,7 +104,7 @@ async def explain(ctx: MtgContext, *, thing: Optional[str]) -> None:
             {}
         ),
         'prices': (
-            f"""
```
bakert, line 98: I think if we lose the `f` here the vars in curly braces in this string will no longer be interpolated.

discordbot/commands/explain.py
```diff
@@ -71,6 +71,15 @@ async def explain(ctx: MtgContext, *, thing: Optional[str]) -> None:
                 'Current League': fetcher.decksite_url('/league/current/')
             }
         ),
+        'netdecking': (
+            """
+            If you are having trouble coming up with an original decklist, feel free to start by learning from the best!
+            You can find the best tournament winning decks in the link below. Sort by stars to get decks that won tournaments or did well in tournaments and leagues.
+            """,
+            {
+                'Decks': fetcher.decksite_url('/decks/'),
```
bakert, line 80: If we do go with the shorter text it might be nice to call this "Decklists" not "Decks"?

bakert:
In general it's a bad idea to submit untested changes – even if the PR "looks right" there's too many little things that escape a human's notice that a machine will not cope with.

discordbot/commands/explain.py
silasary: Yep

discordbot/commands/explain.py
kzcheng: Got it.

discordbot/commands/explain.py
kzcheng: Got it.

kzcheng:
See #7642 for new pull request.

PR #7888, opened 2020-09-28
Always send a rotation hype message in response to the hype comamnd.

bakert:
Even if nothing interesting has happened.

(tbh the skipping of reports when nothing has happened is confusing and should
probably just be removed - #general is buy enough now it's never just the
bot talking to itself and if it is does it matter?)


bakert:
Abandoning this. There is a ticket open to rethink the hype command.

PR #8117, opened 2020-11-24
Use h2 on achievements not bold.

bakert:


bakert:
This was a mostly-spurious commit to get something onto prod but I see it's already there.

PR #8345, opened 2021-02-19
Fix bug in refactor

bakert:


bakert:
I misunderstood what was bugging out here. No problem with refactor just with my text.

PR #8351, opened 2021-02-20
Don't show an incorrect date for PD500

### LINKED ISSUE #8337: site should not show an incorrect date for pd500 if we don't know the length of the season yet
Reported on Discord by silasary#0001

silasary:
Actually fix #8337, and do so in a way that doesn't require null checks.

### DIFF
--- a/magic/tournaments.py
+++ b/magic/tournaments.py
@@ -80,14 +80,15 @@ def get_all_next_tournament_dates(start: datetime.datetime, index: int = 0) -> L
 
 # Note: this may be in the past. It always gives the date for the current season.
 # Note: if the start date of next season is not known then the date of the PD 500 cannot be known and in such a case this return None.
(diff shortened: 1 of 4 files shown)


magic/tournaments.py
```diff
@@ -80,14 +80,15 @@ def get_all_next_tournament_dates(start: datetime.datetime, index: int = 0) -> L
 
 # Note: this may be in the past. It always gives the date for the current season.
 # Note: if the start date of next season is not known then the date of the PD 500 cannot be known and in such a case this return None.
-def pd500_date() -> Optional[datetime.datetime]:
+def pd500_date() -> datetime.datetime:
+    if seasons.next_rotation_ex().codename == '???':
+        return datetime.datetime(1970, 1,1)
```
bakert: Gross!

PR #8418, opened 2021-03-05
Fix homepage

silasary:


decksite/views/home.py
```diff
@@ -45,7 +45,7 @@ def setup_decks(self, decks: List[Deck]) -> None:
                 tournament_id = d.competition_id
             if d.source_name == 'League' and league_id is None:
                 league_id = d.competition_id
-            if d.competition_id is not None and d.competition_id == tournament_id and d.finish <= 8:
+            if d.competition_id is not None and d.competition_id == tournament_id and d.finish is not None and d.finish <= 8:
```
bakert, line 48: What changed to make this necessary now?

PR #8607, opened 2021-04-27
Update card_aliases.tsv

SantiRivera92:


mvousden:
Looks like your `stp` entry is using a space instead of a tab. Also, welcome to GH!

PR #8824, opened 2021-06-22
Alias for Mtenda Lion.

Kalioc:
Because it comes up a lot.


bakert:
@Kalioc card_aliases.tsv is that rarest of beasts, a tab-delimited file – I updated the PR to reflect that.

PR #8915, opened 2021-07-21
Upgrade to Flask 2.0

silasary:


silasary:
Note that I can't actually upgrade to Flask 2.X, because we're waiting on https://github.com/pallets/flask/pull/4094 and https://github.com/pallets/flask/pull/4105 to be released

PR #9024, opened 2021-08-25
Added Dreadrise search

Wizzerinus:
idk if this is needed but why not

if youre pulling this set `dreadrise_url` configuration key to `http://localhost:3001` on production to reduce the network load

### DIFF
--- a/discordbot/command.py
+++ b/discordbot/command.py
@@ -257,3 +257,6 @@ class MtgContext(commands.Context):
     async def post_cards(self, cards: List[Card], replying_to: Optional[Member] = None, additional_text: str = '') -> None:
         # this feels awkward, but shrug
         await post_cards(self.bot, cards, self.channel, replying_to, additional_text)
+
+    async def post_no_matches(self) -> None:
+        await post_no_cards(self.channel)
--- a/discordbot/command.py
+++ b/discordbot/command.py
@@ -139,7 +139,7 @@ async def post_cards(
         additional_text: str = '',
 ) -> None:
     if len(cards) == 0:
-        await post_no_cards(channel, replying_to)
+        await post_nothing(channel, replying_to)
         return
     not_pd = configuration.get_list('not_pd')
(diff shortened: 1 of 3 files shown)


discordbot/command.py
```diff
@@ -257,3 +257,6 @@ async def single_card_text(self, c: Card, f: Callable, show_legality: bool = Tru
     async def post_cards(self, cards: List[Card], replying_to: Optional[Member] = None, additional_text: str = '') -> None:
         # this feels awkward, but shrug
         await post_cards(self.bot, cards, self.channel, replying_to, additional_text)
+
+    async def post_no_matches(self) -> None:
+        await post_no_cards(self.channel)
```
bakert: What does this do? Sounds like `post_no_cards` should be renamed to `post_nothing`?

discordbot/commands/drc.py
```diff
@@ -0,0 +1,109 @@
+from discord.ext import commands
```
bakert, line 1: This is cool.

discordbot/command.py
Wizzerinus: post_no_cards was a function before but sure i can do that

discordbot/commands/drc.py
Wizzerinus: Copied from the scryfall script

bakert: Sorry that comment was meant for the whole file. I am just appreciating your addition :)

discordbot/command.py
bakert: What I am suggesting is that `post_no_cards` be renamed to `post_nothing` and that each callsite for `post_no_cards` and `post_no_matches` be replaced by a call to `post_nothing`.

Wizzerinus: done

discordbot/commands/drc.py
Wizzerinus: ah haha got you

Issue #9027, opened 2021-08-27
dreadrise search errors don't say anything in the chat


vorpal-buildbot:
Reported on Discord by Wizzerinus#5217

Wizzerinus:
fixed in #9062

PR #9062, opened 2021-09-03
fixed #9060 and #9027, now with better code quality

### LINKED ISSUE #9060: drc should use asynchronous fetching, only display 4 when >10 results instead of 10, and shows wrong "and 17779 more"
Reported on Discord by Wizzerinus#5217

Wizzerinus:
mypy hates this because apparently if
```
x is of type Union[Tuple[A,B], Tuple[C, None]]
B is never falsy
and you do if (x[1])
then it thinks x[0] is of type Union[A, C] and not A
also it thinks Tuple[int, []] is not of type Tuple[int, List[str]]
```
but some parts of this code are bad and i don't know what would be the optimal way to fix them. if you know lmk
also while i'm at this i think it's worth making scryfall search asynchronous so making a request while scryfall is under load doesnt hang the entire bot


discordbot/commands/drc.py
```diff
@@ -82,29 +96,32 @@ async def matchups(ctx: MtgContext, *, args: str) -> None:
         domain=link_domain, length=output['length'], wr=output['winrate'], url=output['compress'])
     await ctx.send(ans)
 
-def search_dreadrise(query: str) -> Tuple[int, List[str]]:
+async def search_dreadrise(query: str) -> Union[Tuple[str, None], Tuple[int, List[str]]]:
     """Returns a tuple. First member is an integer indicating how many cards match the query total,
        second member is a list of card names up to the maximum that could be fetched in a timely fashion."""
```
bakert: I think rather than making this convoluted Union type the right answer here is to return three values.

    how_many, cards, error = await search_dreadrise(query)
    if error:
        # handle the error
        return
    # proceed to handle the non-error case
    
Then your type is just:

    Tuple[int, List[str], str]

or maybe:

    Tuple[int, List[str], Optional[str]]
    
if you want to return `None` and not the empty string for "no error". I think it's fine/good to return a `how_many` of 0 and and an empty list of cards in the error case.

discordbot/commands/drc.py
```diff
@@ -82,29 +96,32 @@ async def matchups(ctx: MtgContext, *, args: str) -> None:
...
 
     query = fetch_tools.escape(query)
     url1 = f'{domain}/cards/find?q={query}+output:resultcount'
+    count_txt = await fetch_tools.fetch_async(url1)
     try:
-        count = int(fetch_tools.fetch(url1))
+        count = int(count_txt)
     except ValueError:
-        return 0, []
+        return count_txt, None  # this sucks but i need 2 values to unpack
```
bakert: This would then be `return 0, [], count_txt` or `return 0, [], f'Failed to get count from {count_txt}'`

discordbot/commands/drc.py
```diff
@@ -82,29 +96,32 @@ async def matchups(ctx: MtgContext, *, args: str) -> None:
...
-def search_dreadrise(query: str) -> Tuple[int, List[str]]:
+async def search_dreadrise(query: str) -> Union[Tuple[str, None], Tuple[int, List[str]]]:
     """Returns a tuple. First member is an integer indicating how many cards match the query total,
        second member is a list of card names up to the maximum that could be fetched in a timely fashion."""
     if not query:
         return 0, []
 
     query = fetch_tools.escape(query)
     url1 = f'{domain}/cards/find?q={query}+output:resultcount'
+    count_txt = await fetch_tools.fetch_async(url1)
```
bakert: Our "normal" usage of `fetch_tools` is to create a wrapper function in `fetcher` that uses the lower-level `fetch_tools` functions. So you'd make something like:

    def dreadrise_card_count(query: str) -> int:
        domain = configuration.get_str('dreadrise_url')
        query = fetch_tools.escape(query)
        url = f'{domain}/cards/find?q={query}+output:resultcount'
        txt = await fetch_tools.fetch_async(url)
        try:
            n = int(txt)
        except ValueError as e:
            raise FetchExecption(f'Unable to get card count with query `{query}`') from e
            
and then this whole code block would be replaced by:

    n = fetcher.dreadrise_card_count(query)
    
Notice that when you break it apart like this into well-named functions that do a single thing you then see issues like needing to have vars named `url1` and `url2` melt away.

This entire function becomes something like:

    if not query:
        return 0, [], 'no query'
    cards_fetching = DEFAULT_CARDS_SHOWN if count > MAX_CARDS_SHOWN else MAX_CARDS_SHOWN
    n = fetcher.dreadrise_card_count(query)
    cs = fetcher.dreadrise_card_search(query, True)
    if len(cs) == cards_fetching:
        return n, cs, None
    cs += fetcher.dreadrise_card_search(query, False)
        return n, cs, None

bakert (changes requested):
Some thoughts on your type issue, plus a refactoring that I think makes a lot of sense but is not a blocker.

discordbot/commands/drc.py
Wizzerinus: i like this approach, will do

discordbot/commands/drc.py
Wizzerinus: interesting, i can do that. i don't know how async works in python, but i know dreadrise queries can take a while if they have `sf` or `url` clauses, so the `fetcher.dreadrise_card_whatever` should be async too right? otherwise the bot gets blocked

bakert: Yeah, make them async if they are potentially long-running.

See `all_cards_async` for an example in fetcher that's already async.

Wizzerinus:
addressed all the reviews. not sure if drc actually works tho, new version of aiohttp is broken it seems

bakert:
The "test" failure is the "RG" issue from Tolovar here, but the linter complaints are genuine:

```./magic/fetcher.py:237:17: Q000 Double quotes found but single quotes preferred
./magic/fetcher.py:237:41: Q000 Double quotes found but single quotes preferred
./magic/fetcher.py:238:17: Q000 Double quotes found but single quotes preferred
./magic/fetcher.py:239:55: E261 at least two spaces before inline comment
```

You can run the linter locally with `dev.py lint`

discordbot/commands/drc.py
```diff
@@ -36,11 +42,15 @@ async def decks(ctx: MtgContext, *, args: str) -> None:
     url = f'{domain}/decks/find?q={q}+output:json_api&page_size={MAX_DECKS_SHOWN}'
 
     try:
-        output = fetch_tools.fetch_json(url)
-    except fetch_tools.FetchException:
+        output = await fetch_tools.fetch_json_async(url)
```
bakert: It would be nicer to push these into `fetcher` too.

Wizzerinus: i guess, i don't like the duplicated code here. i guess i could add a lower-level `fetcher.dreadrise_json_from_url` and then `fetcher.dreadrise_matchups` and `fetcher.dreadrise_decks` or something?

bakert: Makes sense to me. In general `fetch_tools` is meant to be low-level plumbing not something you call directly from logic.

Wizzerinus:
by the way, can dreadrise be updated before pulling this? in particular these two commits. (the bottom one is a blocker, the top one is not but is nice to have)
image

magic/fetcher.py
```diff
@@ -212,6 +212,61 @@ def get_frontside(scr_card: Dict) -> str:
...
+    domain = configuration.get_str('dreadrise_url')
+    query = fetch_tools.escape(query)
+    url = f'{domain}/cards/find?q={query}+output:resultcount'
+    count_txt = await fetch_tools.fetch_async(url)
+    try:
+        return int(count_txt), None  # if this fails, the query errored
+    except ValueError:
+        return -1, count_txt
+
+async def dreadrise_search_cards(query: str, page_size: int = 60, pd_mode: Literal[1, 0, -1] = 0) -> List[str]:
```
bakert, line 229: It might be better to represent the tristate `pd_mode` as `{True,False,None}`? Not important.

bakert (approved):
Looks good.

PR #9191, opened 2021-09-28
Allow 'penny' as well as 'pd' for format in search

### LINKED ISSUE #9187: `-f:penny` does not behave as expected

bakert:
Fixes #9187.

### DIFF
--- a/find/search.py
+++ b/find/search.py
@@ -267,7 +267,7 @@ def set_where(name: str) -> str:
     return '(c.id IN (SELECT card_id FROM printing WHERE set_id IN (SELECT id FROM `set` WHERE name = {name} OR code = {name})))'.format(name=sqlescape(name))
 
 def format_where(term: str) -> str:
-    if term == 'pd' or term == 'Penny Dreadful':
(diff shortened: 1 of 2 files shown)


find/search.py
```diff
@@ -267,7 +267,7 @@ def set_where(name: str) -> str:
     return '(c.id IN (SELECT card_id FROM printing WHERE set_id IN (SELECT id FROM `set` WHERE name = {name} OR code = {name})))'.format(name=sqlescape(name))
 
 def format_where(term: str) -> str:
-    if term == 'pd' or term == 'Penny Dreadful':
+    if term == 'pd' or term.startswith('penny'):
```
silasary, line 270: Doesn't this cause issues with `f:"Penny Dreadful AER"`?

bakert: Hm, yes, it does. I hadn't really realized or had forgotten that we supported that but I guess we do. I don't suppose anyone is using it though. We should introduce a shorter syntax for that. `pd16`/`penny16`? Dreadrise uses `pds16` so we could roll with that for compatibility.

PR #9390, opened 2021-11-08
Fix isort job

silasary:


dev.py
```diff
@@ -300,7 +301,6 @@ def check(argv: List[str]) -> None:
     do_check(argv)
 
 def do_check(argv: List[str]) -> None:
-    do_sort(False)
```
bakert, line 303: Why doesn't check get to `do_sort` any more?

silasary: It's time consuming and redundant. 
Imports are automatically fixed by the CI in prs like #9346

PR #9404, opened 2021-11-11
Add a lockfile to rotation

### LINKED ISSUE #9208: We somehow managed to cache half of Run_001.txt during rotation causing erroneous eliminated messages throughout rotation -- we should make that impossible somehow

silasary:
Fixes #9208


discordbot/bot.py
```diff
@@ -326,7 +327,9 @@ async def background_task_rotation_hype(self) -> None:
         while self.is_ready():
             until_rotation = seasons.next_rotation() - dtutil.now()
             last_run_time = rotation.last_run_time()
-            if until_rotation < datetime.timedelta(7) and last_run_time is not None:
+            if os.path.exists('.rotation.lock'):
+                timer = 10
```
bakert, line 331: Exciting. But what does this _do_?

silasary: It schedules the task to try again in 10 seconds, rather than reading files right now

bakert: Oh I see, we already have a `timer` var that we use lower down. I thought it was doing something magical upon assignment! Needed to hit the show more context button on the gihub diff :)

PR #9410, opened 2021-11-12
Warn you on the signup page if you are already signed up

bakert:


decksite/controllers/league.py
```diff
@@ -27,10 +27,13 @@ def current_league() -> wrappers.Response:
 
 @APP.route('/signup/')
 @auth.load_person
-def signup(form: Optional[SignUpForm] = None) -> str:
+@fill_cookies('deck_id')
+def signup(form: Optional[SignUpForm] = None, deck_id: Optional[int] = None) -> str:
+    print(deck_id)
```
silasary, line 32: Print() shouldn't be used

bakert: Oops

PR #9448, opened 2021-11-22
4x the load of /rotation/ with memoization

bakert:


silasary:
How does this interact with rollover?

PR #9505, opened 2021-12-03
Added a deck scraper API

Wizzerinus:
Don't exactly know if I made it in a correct way, but now there's a deck scraper API (under `/api/scraper`) which shows all finished decks from all seasons, which were updated (either finished or had their archetype changed) after a certain moment. This lets me do the daily Dreadrise update faster, not having to pull from the entire season, while also letting me automatically catch updates from older seasons (when a new archetype is created).


decksite/controllers/api.py
```diff
@@ -143,6 +143,40 @@ def decks_api() -> Response:
     resp.set_cookie('page_size', str(page_size))
     return resp
 
+@APP.route('/api/scraper/')
```
silasary: I'd much rather you use the modern API stuff provided by [restx](https://flask-restx.readthedocs.io/en/latest/) (See /api/decks/ as an example).
I've been slowly going through and porting all of the API endpoints, and adding new old-style endpoints is not something I want to be doing

silasary (changes requested):

decksite/controllers/api.py
Wizzerinus: I think I made this work now

Wizzerinus:
Some notes
- the DECK model had OMW as number, but the database outputs `52%` or similar, which can't be converted into number by restx. Changed this, but could be wrong
- For some reason requesting `/api/scraper/?deckType=all&page=10000&pageSize=3&lastTimestamp=1038200000&seasonId=all` (page size 3 for testing purposes) says "59000 decks were found" (which is way less than it should be), but also outputs 0 decks (even though 30000 < 59000). I guess this shouldn't be used for big requests then, but idk if it will persist in real usage
- with the same request, decks on high pages (5500) have no `last_archetype_change`. I guess that's because archetyping wasn't fully implemented back then, but not totally sure (clarification would be good)
- after adding MAX, the query became significantly slower for some reason, (though I guess I also decreased lastTimestamp. thats probably why)

decksite/data/query.py
```diff
@@ -237,3 +237,6 @@ def card_search_where(q: str) -> str:
 
 def tournament_only_clause() -> str:
     return "ct.name = 'Gatherling'"
+
+def deck_scraper(ts: int) -> str:
+    return f') AND d.finish AND (q.changed_date > {ts} OR d.updated_date > {ts}'
```
bakert: Requiring a value in `d.finish` will only get you tournament decks. I assume you don't want that.

decksite/controllers/api.py
```diff
@@ -143,6 +157,41 @@ def decks_api() -> Response:
     resp.set_cookie('page_size', str(page_size))
     return resp
 
+@APP.api.route('/scraper/')
```
bakert: Let's go with /decks/updated/. All the other API endpoints are named after the thing they return.

decksite/controllers/api.py
```diff
@@ -143,6 +157,41 @@ def decks_api() -> Response:
...
+    def get(self) -> Dict[str, Any]:
+        """
+        Grab a slice of finished sorted decks last updated after a certain point.
+        Input:
+            {
+                'sortBy': <str>,
+                'sortOrder': <'ASC'|'DESC'>,
+                'page': <int>,
+                'pageSize': <int>,
+                'lastTimestamp': <int>,
```
bakert: I like `since` or `sinceTimestamp` if you must for this.

`/decks/updated/?since=9393938229`

Nice.

decksite/controllers/api.py
```diff
@@ -143,6 +157,41 @@ def decks_api() -> Response:
...
+            }
+        Output:
+            {
+                'page': <int>,
+                'objects': [<deck>],
+                'total: <int>
+            }
+        """
+        season = seasons.season_id(str(request.args.get('seasonId')), None)
+        order_by = 'GREATEST(cache.active_date, q.changed_date) DESC'
```
bakert: Do you need to order by this? It might be expensive and it seems unnecessary if you're already providing `since`?

decksite/data/query.py
```diff
@@ -237,3 +237,6 @@ def card_search_where(q: str) -> str:
 
 def tournament_only_clause() -> str:
     return "ct.name = 'Gatherling'"
+
+def deck_scraper(ts: int) -> str:
```
bakert: `decks_since`?

decksite/data/query.py
bakert: This should be a complete piece of logic not a partial with a hanging paren. The hanging paren requires the caller to have altogether too much knowledge of what this func does.

```
return 'd.finish AND (q.changed_date > {ts} OR d.updated_date > {tr})'
```

Then in the caller:

```
where = query.decks_where(…) + ' AND ' + query.decks_since(…)
```

bakert (changes requested):
Looks broadly ok. I do wonder if this should be an optional field you can pass to /api/decks rather than an entirely new endpoint. Some comments on the execution.

bakert:
`deck_archetype_change` was only added a few seasons ago.

bakert:
> For some reason requesting /api/scraper/?deckType=all&page=10000&pageSize=3&lastTimestamp=1038200000&seasonId=all (page size 3 for testing purposes) says "59000 decks were found" (which is way less than it should be), but also outputs 0 decks (even though 30000 < 59000). I guess this shouldn't be used for big requests then, but idk if it will persist in real usage

You should figure out what is going on here before making this live.

decksite/data/query.py
Wizzerinus: That's what it initially was, but there are a bunch of hanging ORs at the end of the decks_where request, so it just ignored the OR. The tournament bit is interesting, and explains the 59000 deck issue, thank you.

decksite/controllers/api.py
Wizzerinus: Sounds good

decksite/controllers/api.py
Wizzerinus: sure ill do that

decksite/controllers/api.py
Wizzerinus: that might explain the bad speed, yea this probably makes sense

Wizzerinus:
yea that's why i made it a draft

decksite/data/query.py
bakert: You can use parens to work around this:

```
where = '(' + query.decks_where(…) + ') AND (' + query.decks_since(…) + ')'
```

decksite/controllers/api.py
Wizzerinus: If I don't add `order_by` at all, it will still order in a way consistent between requests right? so the pagination actually would work

bakert: Good point. Maybe try `ORDER BY d.id` and compare the performance to this?

silasary (approved):

decksite/controllers/api.py
```diff
@@ -40,32 +40,40 @@
...
-    'colors': fields.List(fields.String()),
-    'omw': fields.Integer(),
-    'season_id': fields.Integer(),
-    'maindeck': fields.List(fields.Nested(DECK_ENTRY)),
-    'sideboard': fields.List(fields.Nested(DECK_ENTRY)),
-    'url': fields.String(),
-    'source_name': fields.String(),
-    'competition_type_name': fields.String(),
-})
+def get_deck_model(include_last_update: bool) -> Any:
```
bakert: How expensive is including last update? It might be nice just to always send it to avoid this boolean and extra variables like DECK_WITH_ARCH_CHANGE etc. Failing that how about we send it but send it empty?

decksite/controllers/api.py
```diff
@@ -83,6 +91,12 @@
     'decks': fields.List(fields.Nested(DECK)),
 })
 
+MULTIPLE_DECKS = APP.api.model('MultipleDecks', {
```
bakert: This could just be called DECKS probably.

decksite/controllers/api.py
```diff
@@ -143,6 +157,42 @@ def decks_api() -> Response:
     resp.set_cookie('page_size', str(page_size))
     return resp
 
+@APP.api.route('/updated/')
```
bakert: I thought we were going with /decks/updated? We really want a noun in there.

decksite/data/deck.py
```diff
@@ -26,8 +26,9 @@ def load_deck(deck_id: int) -> Deck:
 
 def load_decks_count(where: str = 'TRUE',
                      having: str = 'TRUE',
-                     season_id: Optional[Union[str, int]] = None) -> int:
-    columns = 'COUNT(*) AS n'
+                     season_id: Optional[Union[str, int]] = None,
+                     distinct: bool = False) -> int:
```
bakert: This should be unnecessary. Is there ever a time we call `load_decks_count` and we want to count decks more than once? You can just put the DISTINCT in every time.

bakert (commented):
Looks good to me. I have a few stylistic points that can maybe make things a bit neater.

bakert:
I fixed the unit test failure you are seeing here a few days ago. If you rebase on master it will pass.

decksite/controllers/api.py
Wizzerinus: It's slightly expensive because you need to join and find a MAX on it. It doesn't seem to create a huge overhead though, so I could add it to the decks requests as well. I could also just use the same model, but send nulls when loading from /api/decks. That last approach seems the best

decksite/controllers/api.py
Wizzerinus: oops

decksite/data/deck.py
Wizzerinus: it's required to be DISTINCT only if you join with a table which contains multiple entries per deck, which right now is only `deck_archetype_change`. Seems like an unnecessary overhead to add.

Wizzerinus:
I did a rebase, idk how it works so I guess I'll just pray I didn't break everything. also added all the things you were talking about

decksite/data/deck.py
bakert: My point is that `DISTINCT` always gets the correct result even when it's not strictly necessary. So we can do away with the cognitive overhead of the additional param at the expense of whatever it does to the query. I'm also 90% convinced that it will not affect query speed more than the tiniest bit. So we can essentially do away with the cognitive overhead of the additional param at no cost.

Wizzerinus: okay I made it always distinct

bakert (approved):
Ship it!

PR #9610, opened 2022-01-09
Updated /dreadrise command

Wizzerinus:
recently (today) dreadrise was updated, so I streamlined the APIs, so this PR rewrites the /dreadrise command to be compatible with new APIs, as well as fixing some Interactions-related issues

### DIFF
--- a/discordbot/command.py
+++ b/discordbot/command.py
@@ -336,7 +336,7 @@ class MtgMixin:
         await post_cards(self.bot, cards, self, replying_to, additional_text)
 
     async def post_nothing(self: SendableContext) -> None:
-        await post_nothing(self.channel)
+        await post_nothing(self)
 
 
 InteractionContext.__bases__ += (MtgMixin,)
--- a/discordbot/commands/drc.py
+++ b/discordbot/commands/drc.py
@@ -2,6 +2,7 @@ from typing import Dict, Optional
 
 from dis_snek.models.application_commands import OptionTypes, slash_command, slash_option
 from dis_snek.models.discord_objects.embed import Embed
+from urllib import parse
 
 from discordbot.command import DEFAULT_CARDS_SHOWN, MAX_CARDS_SHOWN, MtgContext
 from magic import fetcher, oracle
@@ -17,7 +18,7 @@ def format_deck(x: Dict) -> Dict:
     return {
         'name': '{name} [{src}, {wins}-{losses}]'.format(
             name=x['deck']['name'], wins=x['deck']['wins'], losses=x['deck']['losses'], src=x['competition']['name']),
-        'value': '[A {arch} deck by {author} ({format})]({domain}/decks/{id})'.format(
+        'value': '[{arch} deck by {author} ({format})]({domain}/decks/{id})'.format(
             arch=x['tags'][0]['name'], author=x['author']['nickname'],
             format=x['format'], id=x['deck']['deck_id'], domain=link_domain),
     }
@@ -39,9 +40,9 @@ async def drc(ctx: MtgContext, query: str) -> None:
     card_array = [x['name'] for x in card_data['sample']]
     count = card_data['matches']
     if count < MAX_CARDS_SHOWN:
-        card_data2 = await fetcher.dreadrise_search_cards(query, MAX_CARDS_SHOWN - card_data['matches'], -1)
-        card_array += [x['name'] for x in card_data2['sample']]
-        count += card_data2['matches']
+        card_data_pd_illegal = await fetcher.dreadrise_search_cards(query, MAX_CARDS_SHOWN - card_data['matches'], -1)
+        card_array += [x['name'] for x in card_data_pd_illegal['sample']]
+        count += card_data_pd_illegal['matches']
 
     cbn = oracle.cards_by_name()
     cards = [cbn[name] for name in card_array if cbn.get(name) is not None]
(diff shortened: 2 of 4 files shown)


discordbot/command.py
```diff
@@ -336,7 +336,7 @@ async def post_cards(self: SendableContext, cards: List[Card], replying_to: Opti
         await post_cards(self.bot, cards, self, replying_to, additional_text)
 
     async def post_nothing(self: SendableContext) -> None:
-        await post_nothing(self.channel)
+        await post_nothing(self)
```
bakert, line 339: You haven't really changed anything here – but isn't this an infinite loop? When you call `post_nothing` you are calling it on a context and not this command. I think. But clearly I don't really understand what this does :)

discordbot/commands/drc.py
```diff
@@ -15,10 +15,11 @@
 def format_deck(x: Dict) -> Dict:
     """Formats a deck object. Returns a dictionary with name and value."""
     return {
-        'name': '{name} [{src} {wins}-{losses}]'.format(
-            name=x['name'], wins=x['record']['wins'], losses=x['record']['losses'], src=x['source']),
-        'value': '[A {arch} deck by {author} ({format})]({domain}/decks/single/{id})'.format(
-            arch=x['archetype'], author=x['author'], format=x['printed_format'], id=x['id'], domain=link_domain),
+        'name': '{name} [{src}, {wins}-{losses}]'.format(
+            name=x['deck']['name'], wins=x['deck']['wins'], losses=x['deck']['losses'], src=x['competition']['name']),
+        'value': '[A {arch} deck by {author} ({format})]({domain}/decks/{id})'.format(
```
bakert: I'd cut the "A" here to avoid grammatical calamity ("A Elves deck") when the archetype name begins with a vowel ("Elves deck by bakert99 …").

discordbot/commands/drc.py
```diff
@@ -15,10 +15,11 @@
...
+            arch=x['tags'][0]['name'], author=x['author']['nickname'],
```
bakert, line 22: Is there guaranteed to be at least one tag always? Might get an IndexError here otherwise.

discordbot/commands/drc.py
```diff
@@ -29,70 +30,83 @@ def format_deck(x: Dict) -> Dict:
...
         return
 
-    cards_shown = DEFAULT_CARDS_SHOWN if count > MAX_CARDS_SHOWN else count
-    cardnames = await fetcher.dreadrise_search_cards(query, cards_shown, 1)
-    if len(cardnames) < cards_shown:
-        cardnames += await fetcher.dreadrise_search_cards(query, cards_shown - len(cardnames), -1)
+    card_array = [x['name'] for x in card_data['sample']]
+    count = card_data['matches']
+    if count < MAX_CARDS_SHOWN:
+        card_data2 = await fetcher.dreadrise_search_cards(query, MAX_CARDS_SHOWN - card_data['matches'], -1)
```
bakert: `var_name2` is a bit of a code smell that might imply breaking this up into more than one func is a good idea.

discordbot/commands/drc.py
```diff
@@ -29,70 +30,83 @@ def format_deck(x: Dict) -> Dict:
 @slash_option('query', 'search query', OptionTypes.STRING)
 async def drc(ctx: MtgContext, query: str) -> None:
     """Card search using Dreadrise."""
-    count, error = await fetcher.dreadrise_count_cards(query)
-    if error:
-        await ctx.send(f'Search error: `{error}`')
+    await ctx.defer()
```
bakert, line 34: What does this do?

discordbot/commands/drc.py
```diff
@@ -29,70 +30,83 @@ def format_deck(x: Dict) -> Dict:
...
-    ans = '{length} matches found. Winrate: {wr}%\n{domain}/minimize/{url}'.format(
-        domain=link_domain, length=count, wr=output['winrate'], url=output['compress'])
+    ans = '{length} matches found. Winrate: {wr}%\n{domain}{path}?q1={q1}&q2={q2}{pd}'.format(
+        domain=link_domain,
+        length=data['matches'],
+        wr=data['winrate'],
+        path='/deck-search/matchups',
+        pd='&dist=penny_dreadful',
+        q1=q1.replace(' ', '%20'),
+        q2=q2.replace(' ', '%20'),
```
bakert: You want to `urllib.parse.quote` here rather than this simple replace. This defends you against (for example) input containing an ampersand ('&') char.

bakert (changes requested):
Looks good as far as I can follow it. A couple little things to maybe fix up.

discordbot/command.py
Wizzerinus: this changes basically, you either send to channel which gives you "interaction failed", or sends it to the interaction. also, one is a method, other is a function

discordbot/commands/drc.py
Wizzerinus: good call

discordbot/commands/drc.py
Wizzerinus: for PD, the tags array is basically always 1-element array with the archetype. unsorted decks aren't scraped

discordbot/commands/drc.py
Wizzerinus: sends the "bot is thinking..." otherwise if request takes more than 3 seconds, you cant reply to it anymore. that happens for matchups i think

discordbot/commands/drc.py
Wizzerinus: good call

discordbot/command.py
bakert: But doesn't `post_nothing` here refer to the function you are currently in? Aka isn't this infinitely recursive?

silasary: Nah, this is (frustratingly) correct, 338 is the definition for `command.MtgMixin.post_nothing`, which then calls `command.post_nothing`

At some point I plan to refactor it to inline the latter

silasary (approved):

bakert (approved):
Ship it!

PR #9818, opened 2022-03-02
Typing changes for inflect update

silasary:
- Typing for inflect
- Handle TestFailedException, so we don't have 40 lines of traceback for a single line of linter warning


bakert:
I did this in 20b9089e

Issue #9819, opened 2022-03-02
Active Run decks are visible through the buy button giving all league players perfect information


vorpal-buildbot:
Reported on Discord by Topher#4820

clates:
It's might be as simple as guarding 

this line
https://github.com/PennyDreadfulMTG/Penny-Dreadful-Tools/blob/8ac56bbb5cf693f11c53a21d7bf5a21f64b14ea4/decksite/templates/buttons.mustache#L16

with this guy
https://github.com/PennyDreadfulMTG/Penny-Dreadful-Tools/blob/46b844282aa116233b4c1248965a4ff8f2d07bda/decksite/prepare.py#L90

but I didn't clone or anything; at work rn might check out and PR later this evening.

PR #9822, opened 2022-03-03
ISSUE-9819 - Guarded the Download, Buy, and Decklist source while the…

clates:
… deck is_in_current_run since each exposes the decklist.

Should address #9819 ;


silasary:
Thanks!

silasary (approved):

PR #9823, opened 2022-03-03
ISSUE-9820 - Stopgap for gzipping

clates:
If you can't get around to figuring out why that endpoint isn't zipped, I figure this only applies to a new dev or fresh environment. 

Also noticed that colorama wasn't installed, but was required? Searched the repo and only came across https://github.com/PennyDreadfulMTG/Penny-Dreadful-Tools/pull/8682 . 

Fee free to decline if you don't need the stopgap.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -103,6 +103,7 @@ traceback-with-variables = "*"
 freezegun = "*"
 molter = "*"
 dis-taipan = "*"
+colorama = "*"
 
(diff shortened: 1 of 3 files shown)


bakert:
I think I fixed this properly in https://github.com/PennyDreadfulMTG/Penny-Dreadful-Tools/pull/9941

PR #10778, opened 2022-12-26
Option to run reprime_cache only once a week

silasary:


bakert:
This makes me so sad!

Having your stats lag reality by a week comes close to invalidating them. I
guess that was effectively happening already, though.

silasary:
The actual solution is to batch things so they don't need to lock the entire DB when calculating.  This is just a stopgap

PR #11077, opened 2023-04-06
Preaggregate _card_stats season-by-season instead of in a single query

bakert:
Hopefully this gets us under the lock wait timeout threshold and causes this to start working again.


bakert:
I force-merged this because there is something wrong with the set up of cards db in CI/CD and the smoke tests that rely on it are failing. But my PR has no failures that `master` doesn't have.

Issue #8823, opened 2021-06-21
When folks use emoji that are not in our main font and not in our subset of Symbola we fallback to whatever their browser can do


bakert:
Options:

- Don't care
- Send all emoji in our subset of Symbola, or send the whole of Symbola (helps with Japanese too?)
- Specifically scan deck names for emoi and subset the ones that have been used so far with plan to regularly update preferably automatically
- Replace unsupported emjoi with their name or nothing

bakert:
Example:

https://pennydreadfulmagic.com/decks/140570/

Which shows to me on a Mac as:

![image](https://user-images.githubusercontent.com/1541/122832605-77b6c880-d2a0-11eb-919d-609d0154366a.png)

bakert:
`SELECT name FROM deck WHERE name <> CONVERT(name USING ASCII);` will give you the affected deck names (or maybe `latin2` is better than `ASCII` as our body font probably supports everything in latin2.

I wrote a little program using fonttools to check which codepoints we don't support:

```
#!/usr/bin/env python
from itertools import chain

from fontTools.ttLib import TTFont
from fontTools.unicode import Unicode

from decksite.database import db

def check_chars(font: TTFont, s: str):
    chars = chain.from_iterable([y + (Unicode[y[0]],) for y in x.cmap.items()] for x in font["cmap"].tables)
    points = [char[0] for char in chars]
    result = ""
    for c in s:
        if ord(c) not in points:
            result += c
    return result

concourse = TTFont("/Users/bakert/Downloads/Concourse Basic/Concourse/OpenType TT/Concourse T3 Regular.ttf", 0, allowVID=0, ignoreDecompileErrors=True, fontNumber=-1)
symbola = TTFont("/Users/bakert/Downloads/Symbola/Symbola.otf", 0, allowVID=0, ignoreDecompileErrors=True, fontNumber=-1)

sql = "SELECT name FROM deck WHERE name <> CONVERT(name USING latin2)"
names = db().values(sql)
all_chars = "".join(set("".join(names))) + "𓆘"
print('-------- Looking for: --------')
print(all_chars)
print('-------- Concourse does not have: --------')
print(check_chars(concourse, all_chars))
print('-------- Symbola does not have: --------')
print(check_chars(symbola, all_chars))

concourse.close()
symbola.close()
```

[MAC-TBAKER pd] pipenv run python3 char_finder.py
-------- Looking for: --------
[...]

PR #11710, opened 2023-10-14
Update CSS fonts to cover every symbol used in a deck name so far

bakert:
And add a script that, although a bit rubbish, gives us a fighting chance of being able to do this again in the future when more symbols get added.

Just adding these fonts without subsetting is a complete no-go as the CJK one is 15MB.

See #8823.


bakert:
Killed in favor of #11711 because I don't understand how to add dependencies, or something.

PR #12094, opened 2024-03-22
(Don't really) upgrade to pytest 8.1.1 from 7.x.x (with some fixes) (instead see if CI/CD will pass with pytest 7)

bakert:
- **Allow upgrade to pytest 8 by fixing some funkiness with db and with_test_db**
- **Make smoke test work when run in isolation**
- **Upgrade to pytest 8.1.1 from 7.x.x**


bakert:
Superceded by #12099

PR #12188, opened 2024-04-15
Surface issues with caching of applied rules when sorting decks

bakert:
Sometimes for reasons unknown _applied_rules will not contain the correct
information. Rebuilding _applied_rules from scratch is very expensive.
However /admin/archetypes is already in the business of selecting from a kind
of "live" _applied_rules (just for the set of decks that have not yet been
classified) rather than using the cache table. So we piggy back off that to
find decks/rules that are in the funky state and display it inline in the deck
sorting table.

It's not entirely beautiful but it's practical.


bakert:
The end result is as follows:

image

PR #12428, opened 2024-05-04
Add manabase generator, hidden from the menu for now

bakert:
This is pretty experimental and terrible but I wanted to get it going end
to end. It should become more useful soon.

To achieve this:

  - Require manabase-solver lib (written by me but a separate repo)
  - Add routes for the manabase generator form and processing
  - Refactor SignUpForm into DecklistForm that contains all the shared
    stuff between league signup, deckcheck and manabase.
  - Move that into decksite rather than in decksite.league because it's not
    league specific.
  - Refactor form out of data into decksite as it doesn't touch the db.
  - Fix some lint warnings by moving initialization of all member variables
    into __init__ for the forms.
  - Move the shared text for the three forms into the common ancestor
    instead of duplicating.
  - Create decksite.manabase to hold the wrapper around manabase-solver.
    This is kind of horrible because manabase-solver is kind of horrible but
    improvements are forthcoming.
  - Move template stuff that is now shared between all three forms into
    mustache fragments instead of duplicating.
  - Teach DecklistForm that it doesn't necessarily have to check legality
    (so that ManabaseForm can accept a <60 cards decklist just fine).


bakert:
image

PR #12432, opened 2024-05-04
Update pre-commit with `add-trailing-comma`, as that's the recommended alternative to the now-removed flake8-commas

silasary:
flake8-commas was removed in #12426


.pre-commit-config.yaml
```diff
@@ -27,4 +27,8 @@ repos:
     rev: v3.15.2
     hooks:
     -   id: pyupgrade
-        args: [--py310-plus]
+        args: [--py311-plus]
```
bakert, line 30: Is this dangerous if we're on py310 on prod still?

PR #12999, opened 2024-08-16
Avoid issue creation being rejected by github by truncating overly long values

bakert:


silasary:
Ah, that explains the numbers

bakert:
:D

PR #13117, opened 2024-09-14
Remove types-certifi, trying to avoid a dependabot warning

bakert:
We don't seem to use it, mypy passes without it, and there's a security issue with certifi and dependabot claims it can't upgrade us.

I'm not actually sure it's anything to do with types-certifi but trying this.


bakert:
Acutally this is nonsense

PR #13340, opened 2024-11-18
Revert "Don't clear decksite rotation cache via web, be more direct"

bakert:
This reverts commit d9dc4ab42f30fdc0eb93fa87e437878dc40af8e2.

While this is better it does not work on prod the way it's currently set up to
run in a directory that is not a Penny-Dreadful-Tools checkout :(

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -135,6 +135,7 @@ deny-imports =
     # maintenance can (and does) use anything
 
     rotation_script=analysis
+    rotation_script=decksite
     rotation_script=discordsite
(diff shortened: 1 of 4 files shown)


silasary:
For what it's worth, I don't want rotation script pulling in extra
dependencies if we can avoid it.

bakert:
Yeah I agree. I was getting desperate! I found a way to make it work poking
the web though by spawning a process that calls a maintenance task instead
of running in a thread.

PR #13478, opened 2025-01-08
Dockerfile: use official python and node images

dunn:
Using an unofficial image is potentially risky, and a combined Python/Node isn't necessary here since Node is only a build-time dependency.


bakert:
I guess this makes sense but the docker image is only used for dev. It's not deployed this way. So perhaps it doesn't matter.

What inspired you to make this PR?

dunn:
I thought I would take a stab at some of the open tickets related to the Docker setup, and I assumed it was being used in production so this seemed like a good starting point.

bakert:
Ok sweet. Thanks for taking a look!

Issue #12451, opened 2024-05-05
docker dev setup doesn't spin up logsite


bakert:
I don't think

bakert:
Fixed in commit 3950dc58 (2025-01-08) which added the logsite service to docker-compose.yml with port 5001 binding and all required volume mounts (logsite, logsite_migrations, shared, shared_web, compiled_js).

<sub>Automated backlog sweep; independently verified. Wrongly closed? Please reopen.</sub>

Issue #12450, opened 2024-05-05
docker dev setup instructions should probably tell you how to view local decksite, logsite and db


bakert:

bakert:
Fixed in commit 62bafb3b which updated the README to list all local URLs available after running docker compose up: decksite at port 80, the Adminer database admin panel at port 8080, and logsite at port 5001.

<sub>Automated backlog sweep; independently verified. Wrongly closed? Please reopen.</sub>

PR #13486, opened 2025-01-11
add logsite to docker compose configuration

dunn:
Addresses https://github.com/PennyDreadfulMTG/Penny-Dreadful-Tools/issues/12450 and https://github.com/PennyDreadfulMTG/Penny-Dreadful-Tools/issues/12451

### DIFF
--- a/docker_dev/mysql/01-setup.sql
+++ b/docker_dev/mysql/01-setup.sql
@@ -9,5 +9,4 @@ GRANT CREATE ON *.* TO 'pennydreadful'@'%';
 
 CREATE DATABASE decksite;
 CREATE DATABASE cards;
-CREATE DATABASE pdlogs;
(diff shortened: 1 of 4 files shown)


docker_dev/mysql/01-setup.sql
```diff
@@ -9,5 +9,4 @@ GRANT CREATE ON *.* TO 'pennydreadful'@'%';
 
 CREATE DATABASE decksite;
 CREATE DATABASE cards;
-CREATE DATABASE pdlogs;
```
dunn, line 12: creating an empty db here causes the check in logsite to skip all the migrations: https://github.com/PennyDreadfulMTG/Penny-Dreadful-Tools/blob/master/logsite/__init__.py#L26

PR #13489, opened 2025-01-12
Thousands of errors here daily, full investigation later

bakert:


bakert:
It's almost certainly the missing matches from reverting to the old server where pdlogs is very out of date.

PR #13509, opened 2025-01-15
[docker] simplify dockerfile, add separate assets builder to compose

dunn:
- remove the build stages from the dockerfile; since this is for development, a small image size isn't super important

- move the javascript handling (dependencies and react compilation) to a separate service which can run `npm run watch` and automatically recompile jsx


silasary (approved):
Looks good

PR #13592, opened 2025-03-04
[templates/deckmatches] don't make BYE rows js-clickable

dunn:
~~real HTML anchors are already generated for decks and users; adding the data-href and clickable class has the effect of making BYE entries take the user to the current URL plus False which of course 404s.~~

having `data-href` and the `clickable` class applied to a BYE row has the effect
of making those entries entries take the user to the current URL plus `False`
which of course 404s.

an example: https://pennydreadfulmagic.com/decks/259753/

### DIFF
--- a/decksite/templates/deckmatches.mustache
+++ b/decksite/templates/deckmatches.mustache
@@ -14,7 +14,7 @@
     </thead>
     <tbody>
         {{#matches}}
-            <tr data-href="{{opponent_deck_url}}" class="clickable">
+            <tr {{#opponent_deck_url}}data-href="{{opponent_deck_url}}" class="clickable"{{/opponent_deck_url}}>
                 {{#has_rounds}}
                     <td class="n" data-text="{{round}}">{{display_round}}</td>
                 {{/has_rounds}}


bakert:
Nice. I did not know this happened but it makes sense.

This makes the table behave differently from the other tables on the site. The idea of `.clickable` is that if you're anywhere on a row relating to a deck you might want to click rather than having to be poised over the deck's name. See /league/current for example, where you can click on (say) the OMW value for a deck or in empty space in the row and it'll take you to the deck page. We also do it for cards and people too.

Maybe we can just skip it when the deck id is None rather than all the time? I kind of like `.clickable`.

Somewhat relatedly at some point byes were displayed more nicely and did not say `None` which I believe is the python null value being rendered as a string (yuk!) but I fucked something up in a refactor years ago and have never gone back and figured out what because deck=`None` and opponent=`BYE` does mostly make sense, it's just a little ugly

dunn:
that would also work; i'm not familiar with mustache templating so i couldn't see where to do the logic that would skip non-actual decks.

personally i don't like `clickable` -- i think it breaks the usual web conventions -- but that's not a hill i'm dying on here.

bakert:
Mustache is "logic-less" so you have to set up a var in the template that'll tell you what you want to know In this case we already have `opponent_deck_url` so it'd be something like:

```
<tr {{#opponent_deck_url}}data-opponent-deck-url="{{opponent_deck_url}}" class="clickable"{{/opponent_deck_url}}>
```

For me `.clickable` is like when you choose to make a big rectangle with a menu item in it entirely clickable instead of just the text of the menu option. But it's possible that it's not as analogous to that as I think.

dunn:
(my position would be that if we want that behavior we should just wrap the larger element in an `<a>` rather than use javascript, so that things like right-clicking to 'open link in new tab' etc. work correctly.)

dunn:
updated with your suggestion, thanks!

bakert:
I think the problem with an a tag is that a browser doesn't expect a whole `<tr>` to be wrapped in an `<a>` and that doesn't work how you'd want. I think. But possibly I just messed up the execution.

Thanks for this – it's definitely an improvement regardless of what perfection looks like!

bakert:
I'm going to merge this despite the failing test because the failing test is all about the big oracle wording change where they now say "this card" or "this land" instead of "~" in card templates. I will fix that Soon(tm).

dunn:
> I think the problem with an a tag is that a browser doesn't expect a whole `<tr>` to be wrapped in an `<a>` and that doesn't work how you'd want. I think. But possibly I just messed up the execution.

you might need to have a div inside each td, and link them.

PR #13630, opened 2025-03-11
Put a test back how it should be

bakert:
Curious if this will pass github flows.


bakert:
I did this in another branch now that the API is responding correctly to CI/CD. We might want to do something more robust here in the long-term.

PR #13671, opened 2025-03-22
[download_scryfall_card_image] fix for dfc with version=border_crop

dunn:
previously the code generating the pair of paths for dfcs assumed that `version` would be the extension and would be contained in `filepath`.  however, `version` can also be `border_crop`, so instead of calling a replace on `version` we split the filepath on the extension and insert the a/b strings.


bakert:
I'm gonna be a dick and say this would be nicer as:

```
path, ext = os.path.splitext(filepath)
```

insted of `split` as a list throughout

But thanks for fixing!!

PR #13697, opened 2025-03-26
Account for git-lfs requirement in README, run.py build, and Docker setup

bakert:
We use git-lfs to store font binaries as a reference rather than storing the
files themselves. That way the repo doesn't balloon in size (in particular if
they change).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -59,12 +59,12 @@ jobs:
     - uses: getong/mariadb-action@v1.11
       with:
         mysql root password: bad-password
+    - run: pipenv run init-cards
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip pipenv wheel
         if [ -f Pipfile ]; then pipenv sync --dev; fi
     - uses: PennyDreadfulMTG/setup-linters@main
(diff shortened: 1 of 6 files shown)


dunn:
oh neat, i'd been struggling to get fonts built for the docker environment.  if you want to pull in this commit too, it lets us run dev scripts in docker compose instead of having to install everything on the host system as well: https://github.com/dunn/Penny-Dreadful-Tools/commit/949a4925e4a47bad1594722e3996f5fba1161782

bakert:
This is stalled out because buildfonts requires a working cards db, init_cards requires a working build. So we can't slot buildfonts into buildpy, buildjs, etc.

Perhaps there is another step called "init" that comes after build that does init_cards and creates the fonts.

Manual dev setup can just run two commands.

But docker will need one ENTRYPOINT that does all the necessary, including wait-for-db (which currently fails, actually)

So maybe a little refactor of run.py/build.py/dev.py so they make a little more sense and so the ENTRYPOINT for docker is called somethign like "start" or "setup" and not "run.py --wait-for-db" with the odd callback semantics.

Letting this percolate for a minute and will then either refactor to perfection or hack something in.

PR #13771, opened 2025-04-11
don't hide search box on small screens

dunn:
instead, place it below the header text, and tighten up the vertical spacing

image

### DIFF
--- a/decksite/templates/header.mustache
+++ b/decksite/templates/header.mustache
@@ -30,7 +30,7 @@
     <body>
         <div class="page">
             <header class="season-{{season_id}}">
-              <div>
+              <div class="header-row">
                 <div class="hamburger">☰</div>
                 {{> language_switcher}}
                 <h1><a href="{{home_url}}">Penny Dreadful Magic</a></h1>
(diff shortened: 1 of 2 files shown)


bakert:
You think search is that useful?

dunn:
i think so, yeah

bakert:
Thanks for looking at this sorry to be a nitpicker!

I am willing to give up valllllluuuuuuuable vertical real estate to this (and I am secretly flattered that you like my funny search typeahead) but only if we can make the hamburger menu stay on the same line as the page title even at very narrow iPhone 11 and similar widths.

# Before

Screenshot 2025-04-12 at 17 18 04

# With this PR

Screenshot 2025-04-12 at 17 18 11

dunn:
ah, gotcha.  i'll try to make the 🍔 behave

dunn:
updated, although i'm not sure if it's the most elegant way of doing it

bakert:
Great. If it works it works!

As a consequence of this we've moved the language switcher on wide views. But I don't think it matters.

image

Thanks!

PR #13994, opened 2025-06-26
Change README.md to clear up developer environment setup to be able to log in on the decksite

Tygrak:
Wrote out all the steps in some more detail and added the fact that people need to add a redirect on the discord devsite. Also added a mention that you need to do these steps to be able to log in on the site

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,6 @@ If you plan on running things outside of the containers (eg: dev.py or logsite):
 - Still under the OAuth2 section, add a redirect with the following URI `http://127.0.0.1/authenticate/callback/`
 - Go to the Bot section of your application
 - Reset the token, use it to fill out `token` in `.env`
-- In the root folder of the project copy `config.json.example` to `config.json` and alter the oauth and token values to the ones you put in `.env`
 - Optionally, add the bot to your server with `https://discordapp.com/oauth2/authorize?client_id=<your client id here>&scope=bot`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -67,15 +67,18 @@ If you plan on running things outside of the containers (eg: dev.py or logsite):
...
-- You will want to investigate the various targets in dev.py that acts as a Makefile. Some of these utilities use GitHub's commandline tool, gh: <https://cli.github.com/>
+- Add a new application at <https://discordapp.com/developers/applications/me>
+- Go to the OAuth2 section in your application
+- Reset the client secret
+- In the root folder of the project copy `.env.example` to `.env`
+- In `.env` set `oauth2_client_id` and `oauth2_client_secret` to the client id and client secret found in the OAuth2 section
+- Still under the OAuth2 section, add a redirect with the following URI `http://127.0.0.1/authenticate/callback/`
+- Go to the Bot section of your application
+- Reset the token, use it to fill out `token` in `.env`
+- In the root folder of the project copy `config.json.example` to `config.json` and alter the oauth and token values to the ones you put in `.env`
```
silasary: This is unnecessary, config.json pulls from .env if the key is missing.

bakert: It's kind of awkward that this section is kind of for both docker and non-docker dev and for non-docker dev you probably DO want to do this (but don't care about .env). It's fine to just make docker the Approved Way(tm) though so I'll approve/merge this.

PR #13998, opened 2025-06-29
Added kickoff top8s and victories achievements

Tygrak:
I just think it'd be neat to have them as achievements :)
obrazek
obrazek


bakert:
Cool

PR #14004, opened 2025-06-29
Fixed kick off achievements ignoring kickoffs which have incosistent naming

Tygrak:
Also made the pd 500 and kickoff achievements check competition flag name instead of the competition name
obrazek

(I hope adding another left join into competition_join() won't make it much slower, maybe it can be done better?)


bakert:
The join should be painless. Thanks!

PR #14005, opened 2025-06-30
Never overwrite a main-text symbol with an emoji symbol

bakert:


bakert:
Although this "works" it makes some deck names into only invisible chars which is not good

PR #14713, opened 2026-05-06
Added Prepare layout to layout.py to make the new SOS prepare cards be able to get registered in league

Tygrak:
oh and also it has gay lesbian and queer added to the profanity whitelist which I forgot to make a PR for last year I guess?


Tygrak:
already merged from https://github.com/PennyDreadfulMTG/Penny-Dreadful-Tools/pull/14710

PR #14805, opened 2026-08-21
Warn when card information is stale

bakert:
## Summary

- warn on card command responses when the cards database is more than 48 hours behind Scryfall
- include the warning on both successful lookups and no-match responses
- use the existing one-row Scryfall version timestamp, avoiding network requests and card-table scans

## Testing

- `.venv/bin/ruff check discordbot/command.py discordbot/command_test.py`
- `.venv/bin/mypy discordbot/command.py discordbot/command_test.py`
- `.venv/bin/pytest --confcutdir=discordbot discordbot/command_test.py -q -k 'not results_from_queries and not do_not_choke'` (6 passed; the two excluded existing tests require the local cards MySQL database and Whoosh index)

### DIFF
--- a/discordbot/command.py
+++ b/discordbot/command.py
@@ -15,7 +15,7 @@ from interactions.models import DM, TYPE_MESSAGEABLE_CHANNEL, AutocompleteContex
 
 from discordbot import emoji
 from discordbot.shared import channel_id, guild_id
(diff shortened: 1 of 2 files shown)


bakert:
@Mergifyio refresh

PR #14812, opened 2026-08-21
Harden uv development and CI workflows

bakert:
## Summary
- make local, CI, Docker, and application uv execution consume the committed lockfile with frozen syncs and runs
- consolidate import sorting on Ruff and remove obsolete isort and Pipfile maintenance workflows
- pin uv in CI and Docker, preserve default generated-directory exclusions, and copy uv.lock into the Docker dependency layer

## Validation
- uv lock --check
- uv sync --frozen --dev
- uv run --frozen python dev.py sort
- uv run --frozen python dev.py lint
- uv run --frozen python dev.py mypy
- JavaScript lint and production webpack build
- workflow YAML, TOML, shell, and Python syntax validation

The import-order check completed in 2.8 seconds at about 87 MB RSS without traversing .venv, .git, node_modules, caches, build, or dist. The regenerated uv.lock checksum remained unchanged through every read-only check. Full pytest collection is blocked locally because the configured MySQL user cannot create the missing pdlogs database (error 1044).


bakert:
@mergifyio queue

PR #14917, opened 2026-08-22
Support equivalent card names and canonicalize stored data

bakert:
## Summary

- derive English alternate printed names from Scryfall, including later printings and double-faced cards, while rejecting ambiguous/canonical collisions
- accept either Spider-Man or Omenpaths names for deck entry, league validation, card commands, filters, and search, but store and emit the Scryfall canonical name
- enforce one shared four-copy limit and one deck identity across equivalent names
- redirect alternate-name card-page URLs to the canonical public URL while preserving season and tournament filters
- canonicalize new rotation run files, `rotation_runs`, rotation summaries, and generated `legal_cards.txt` before counting equivalent identities
- provide a Scryfall-derived dry run and collision-safe, idempotent data migration for old `deck_card`, `deck.featured_card`, `rule_card`, and `rotation_runs` values
- create persistent pre-migration backup tables before changing source rows, then refresh affected deck hashes and rotation/cache state
- add SHA-256 values to the rotation publication checklist so GitHub Pages and S3 can be verified as the same generated files
- leave PDBot unchanged

Public decklists remain machine-readable and use canonical Spider-Man names. Validation errors preserve the submitted alias for clarity, for example `Origin of Spider-Man (entered as A Most Helpful Weaver)`.

## Production snapshot rehearsal

Rehearsed against `decksite-2026-08-22.sql.gz` with a card database rebuilt by this branch. The rehearsal used the complete production deck/rule data and every production rotation row belonging to any Scryfall alias identity.

- 680 aliases derived from the rebuilt Oracle data; no hardcoded Omenpaths map
- 2 aliased `deck_card` rows in 2 decks, both `Knife Trick` -> `Pumpkin Bombardment`
- 0 aliased `deck.featured_card` rows
- 0 aliased `rule_card` rows
- 33,927 aliased `rotation_runs` rows: the previously measured 33,117 Omenpaths rows plus 810 older alternate-name rows
- 0 deck, rule, or rotation identity collisions in production
- after apply: zero stored aliases, both deck quantities preserved, both deck hashes matched the normal application calculation, and 460,671 relevant rotation rows plus the complete affected decks were retained in backup tables
- a second audit and apply found zero source changes and completed idempotently

The migration test also covers the harder cases not present in production: alias plus canonical rows, two aliases for one identity, quantity merging, main deck versus sideboard separation, rule semantics, rotation primary-key collisions, dry-run behavior, backups, and reruns.

## Validation

- `uv run pytest -q`: 729 passed, 2 skipped
- `uv run python dev.py lint`
- `uv run mypy .`
- focused migration/rotation suite: 12 passed, 1 pre-existing network test skipped
- pinned Scryfall shadow comparison against `origin/master`: no canonical changes, removals, or ambiguities; 560 aliases retained, 120 added; 910,986 legality rows retained, 80 alias-derived rows added

## Deployment runbook

Take a fresh decksite database snapshot and use a short maintenance window for the migration.

1. Deploy this commit to Columbus consumers of the shared card database. PDBot is not part of this deployment.
2. Rebuild the card database first: `uv run python run.py init-cards --force`.
3. Dry-run the stored-data migration: `uv run python run.py maintenance canonicalize-card-names-audit`. Compare its output with the snapshot counts above and stop if it reports unexpected collisions.
4. Apply it: `uv run python run.py maintenance canonicalize-card-names`. It takes an advisory lock, creates `_canonicalize_card_names_backup_*` tables, and changes the four source locations in one transaction. Hash/cache repair is safe to resume by rerunning the command if a post-commit step is interrupted.
5. Run the audit again and require zero alias rows in every source.
6. Rebuild derived site and search data: `uv run python run.py maintenance reprime-cache` and `uv run python run.py maintenance populate-whoosh-index`.
7. Restart long-lived Decksite and Discord bot processes so their in-memory Oracle maps reload. This does not mean PDBot.
8. Make GitHub Pages and S3 serve the exact same canonical `legal_cards.txt` and current-season legal-list artifact. Verify their SHA-256 hashes match; future rotation issues include the expected hashes automatically.
9. Smoke-test league entry once with each equivalent name, a mixed-name four-copy violation, both card pages/search names, both affected historical decks, and the public legal-card files.

## Rollback

- Any source-write failure rolls back the source transaction automatically.
- The safest rollback after a successful apply is the fresh database snapshot.
- For an immediate targeted rollback during the maintenance window, the `_canonicalize_card_names_backup_deck`, `_deck_card`, `_rule_card`, `_rotation_runs`, and `_mapping` tables contain the complete original affected rows and hashes. Restore those rows, then rerun the cache/search rebuilds and restart processes. Do not use a targeted restore after accepting new deck or rotation writes; use the full snapshot instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,6 +57,8 @@ whoosh_index/
 
 # Scryfall
 /scryfall-default-cards.json
+/scryfall-default-cards.jsonl.gz
+/sets.json
 
 # MTGJSON
 AllCards-x.json
--- a/README.md
+++ b/README.md
@@ -146,6 +146,18 @@ There are various levels of granularity but in general use you want:
 
 Check the dev.py source code for the full set of options including `unit`, `types`, `lint` (covered by `test` above) as well as `functional` (integration tests), `perf` (performance tests). `release` will take you all the way from your committed change to a PR via the tests (needs GitHub's commandline `gh` installed).
 
+### Validating card-import changes
+
+Changes to the Scryfall importer should also be checked with two disposable databases made from exactly the same bulk data:
+
+1. Create a worktree at `origin/master` and place the same `scryfall-default-cards.json` and `sets.json` in both worktree roots. The candidate also accepts Scryfall's compressed `scryfall-default-cards.jsonl.gz` format.
+2. Force an import from the master worktree into a database named `pd_alias_shadow_master` and from the candidate into `pd_alias_shadow_candidate`. Shadow database names deliberately use the `pd_alias_shadow_` prefix to keep them separate from development and production data.
+3. From the candidate worktree, run `uv run python dev.py compare-card-databases pd_alias_shadow_master pd_alias_shadow_candidate`.
+
+The comparison fails if canonical card data was removed or changed, if an existing alias or legality was removed, or if aliases became ambiguous. It permits and reports newly recognized aliases and legality rows derived from those aliases.
+
+After importing current cards, run `uv run pytest magic/omenpaths_test.py rotation_script/rotation_script_test.py` for the focused acceptance cases. They cover normal and double-faced alternate names, mixed-name aggregation, the shared four-copy limit across main deck and sideboard, canonical deck and legal-list output, legacy name matching, and search indexing. The full `uv run pytest` suite remains the final regression gate.
+
 ## Working on React components
 
 - Run logsite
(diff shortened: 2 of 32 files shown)


bakert:
@Mergifyio dequeue

PR #14986, opened 2026-08-24
Tone down the archetypes-to-tag badge

bakert:
The mod/demimod archetypes-to-tag count on the Metagame menu item was a loud red circle. This draws it in the menu's own ink instead, and fixes what was wrong with it besides the color.

| before | after (light) | after (dark) |
|---|---|---|
| loud red circle, count sitting high, floating clear of the icon | acajou disc, snow numerals, hung off the triangle's corner | silver disc, night numerals |

### The count was never centred

`$(".admin").show()` writes an inline `display: inline` onto the badge when it reveals it to a mod. An inline style beats anything the stylesheet says, so flexbox centring could not apply — `position: absolute` then blockified it and `line-height: 1` parked the digits at the top of the box. `line-height` equal to the height survives that and does the centring. Measured against the circle's centre, ink bounding box vs geometric centre:

| count | before | after |
|---|---|---|
| 5 | 22.0% high | 0.4% |
| 80 | 22.3% high | 0.0% |
| 137 | 22.5% high | 0.7% |

### It was not actually a circle

Width grew with the digit count while `border-radius: 50%` followed along, so two digits rendered an oval. Width is now pinned to the height via `min-width`, with padding narrow enough that two digits stay inside it. A three-digit count becomes a short pill rather than spilling out of the circle.

### It was bigger than the icon it belongs to, and detached from it

The `△` glyph draws only 16 x 16px of ink inside its 32px box, while the badge was ~25px sitting 11px clear of it. It is now 12px type in a 20.4px disc hung off the icon's top-right corner.

### The count did not meet AA

Acajou on Spanish gray is **4.40:1**, under the 4.5:1 floor for text this size, and the old `color: white` was a literal rather than a token so it stayed white in dark mode. No gray in the palette carries dark text at AA — Acajou on Isabelline passes at 10.8:1 but is too close to the menu behind it to read as a badge. So the fill inverts:

```css
--badge-background: var(--text);       /* Acajou / Silver */
--badge-text: var(--background);       /* Snow / Night */
```

**11.91:1** in light mode, **12.51:1** in dark — the site's own body-text contrast. No new color: it is the same ink the menu text is already drawn in, and both aliases flip with the palette rather than needing a dark-mode override.

### Narrow screens

Dropped the `max-width: 640px` offsets. The hamburger drawer uses the same 120px menu column as the wide-screen strip, so they only moved the badge back off the icon.

### Testing

Rendered against the real stylesheet at 1000px wide (the badge's own media query means anything narrower silently tests the drawer rules instead), with the inline `display` the JS applies, in both themes, at one, two and three digits. Contrast ratios computed from the palette's OKLCH values and confirmed against sampled pixels.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/f9f40591-dd3b-4677-a0b4-508d26223a45)

### DIFF
--- a/shared_web/static/css/pd.css
+++ b/shared_web/static/css/pd.css
@@ -588,6 +588,10 @@ Color attributes are mostly found with their related structural elements in this
     --tertiary-background: var(--light-gray);
     --tertiary-text: var(--black);
 
+    /* Notification badges. The page colors inverted: the same ink the menu is drawn in, so it stands out against the menu without introducing a color, and the count reads at full text contrast. */
+    --badge-background: var(--text);
+    --badge-text: var(--background);
+
     /* Colors of Magic (avoid using these to represent anything else). */
     --w-mana: #fffcda;
     --u-mana: #ace2fa;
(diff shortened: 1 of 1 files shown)


bakert:
@mergifyio queue

PR #15049, opened 2026-09-01
Determine admin menu permissions from view decorators (#12716)

### LINKED ISSUE #12716: It would be cool to figure out permission_required in admin_menu via reflection looking at the decorator rather than sort-of hardcoded

bakert:
Fixes #12716.

Replace the hardcoded admin-menu name allowlist with the permission_required metadata already attached by admin_required and demimod_required. This keeps menu visibility aligned with the route decorators.

Validation:
- uv run --frozen python dev.py lint — passed
- Focused admin controller tests could not start locally because MariaDB is not running in this manager sandbox; repository CI is the test gate.


bakert:
@Mergifyio queue

PR #15053, opened 2026-09-01
Add --legibility-text-shadow CSS variable for text over images (#12576)

### LINKED ISSUE #12576: Figure out a standard "legibility drop shadow"
This is for anything that appears directly over images.

The site heading has something already.

The search bar probably needs something?

bakert:
## What was wrong

The legibility drop shadow (`0.05rem 0.05rem 8px black`) applied to `header h1` was hardcoded inline with no way to reuse it on other elements that also appear over the site's banner image. There was no standard variable for this pattern.

## What changed

In `shared_web/static/css/pd.css`:
1. Added `--legibility-text-shadow: 0.05rem 0.05rem 8px black` to the `:root` color block (near `--text-over-images`), with a comment explaining it is always dark even in dark mode because it overlays an image.
2. Replaced the hardcoded inline value on `header h1` with `var(--legibility-text-shadow)`.
3. Applied `text-shadow: var(--legibility-text-shadow)` to `header .search`, which also sits over the banner image alongside `header h1`.

## Validation

`uv run --frozen python dev.py lint` — passed. Unit tests were not run (no database available in this environment), but the change is pure CSS with no logic paths to test.

Fixes #12576

### DIFF
--- a/shared_web/static/css/pd.css
+++ b/shared_web/static/css/pd.css
@@ -404,7 +404,6 @@ header h1 {
 header .search {
     flex-grow: 1;
     flex-shrink: 1;
-    text-shadow: var(--legibility-text-shadow);
 }
 
 main, nav {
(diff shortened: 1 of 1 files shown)


bakert:
@mergifyio queue

PR #15054, opened 2026-09-01
Move className propType from DataManager to Table and Grid separately (#12586)

### LINKED ISSUE #12586: Rename className on Grid and on Table and make them separate
the shared propType className on datamanager is really two - one on table that allows with-marginalia, and one on grid that allows metagame-grid, we don't even use them the same way and datamanasger doesn't use them so break them up

bakert:
## What was wrong

`DataManager.propTypes` declared `className` as `PropTypes.oneOf(["", "with-marginalia", "metagame-grid"])`, mixing values that belong to two different child components. `"with-marginalia"` is only valid for `Table` and `"metagame-grid"` is only valid for `Grid`, so the shared base class was conflating them.

## What changed

- **`datamanager.jsx`**: Removed the `className` entry from `DataManager.propTypes`.
- **`table.jsx`**: Added `"className": PropTypes.oneOf(["", "with-marginalia"])` to `Table.propTypes`.
- **`grid.jsx`**: Added `"className": PropTypes.oneOf(["", "metagame-grid"])` to `Grid.propTypes`.

No caller changes were needed — callers already pass `className` under that name to the correct child component.

## Validation

`uv run --frozen python dev.py lint` passed on all three touched files. Unit tests require a live MariaDB instance; the fix is purely static PropTypes metadata (no runtime logic change), so there is no meaningful unit test to add or run.

Fixes #12586

### DIFF
--- a/shared_web/static/js/datamanager.jsx
+++ b/shared_web/static/js/datamanager.jsx
@@ -138,7 +138,6 @@ DataManager.propTypes = {
     "archetypeId": PropTypes.string,
     "baseQuery": PropTypes.string,
     "cardName": PropTypes.string,
-    "className": PropTypes.oneOf(["", "with-marginalia", "metagame-grid"]),
     "competitionId": PropTypes.string,
     "competitionFlagId": PropTypes.string,
     "competitionSeriesId": PropTypes.string,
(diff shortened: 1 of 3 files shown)


bakert:
@mergifyio queue

bakert:
@mergifyio refresh

PR #15057, opened 2026-09-01
Move /api/archetypes2 decoration into a dedicated preparation path (#12590)

### LINKED ISSUE #12590: Consider moving the key_cards and other stuff in /api/archetypes2 into prepare_archetype
Or refactor entirely.

bakert:
## What was wrong

`/api/archetypes2` decorated every result inline with key-card images, match count, and mana-color HTML. The original version of this PR moved that work by adding two related optional arguments to the generic `prepare_archetype` and `prepare_archetypes` functions. That made an endpoint-specific concern part of every archetype preparation call and allowed incomplete argument combinations.

## What changed

- Add a typed, dedicated `prepare_archetypes_for_api` path for the `/api/archetypes2` result shape.
- Move key-card filtering, five-card limiting, Scryfall art URLs, null-safe `num_matches`, and `colors_safe` calculation into that path.
- Keep the generic `prepare_archetype` and `prepare_archetypes` signatures unchanged, and call them after the API-specific decoration.
- Reduce the controller to loading the two inputs and invoking the dedicated preparation function.
- Move the key-card-interest helper alongside the decoration that uses it.

API output and behavior are unchanged.

## Interaction with #15075

This branch is based on current `master`; #15075 is queued but has not merged. I also applied #15075 followed by this commit in a clean integration worktree. Both commits apply without conflicts, the controller retains #15075's current-page key-card scoping, and the focused tests pass on the combined tree.

## Tests

- Add focused coverage for populated and empty archetypes, including filtered/capped `key_cards`, art URLs, null-safe `num_matches`, `colors_safe`, and the full card set used for color calculation.
- `uv run --frozen pytest decksite/prepare_test.py decksite/controllers/api_test.py` — 8 passed
- `uv run --frozen python dev.py unit` — 874 passed, 1 skipped, 94 deselected
- `uv run --frozen python dev.py mypy` — passed (366 files)
- Combined #15075 + #15057 focused tests — 8 passed

Fixes #12590

### DIFF
--- a/decksite/controllers/api.py
+++ b/decksite/controllers/api.py
@@ -16,12 +16,11 @@ from decksite.data import rotation as rot
 from decksite.data import rule as rs
 from decksite.data.achievements import Achievement
 from decksite.data.clauses import DEFAULT_GRID_PAGE_SIZE, DEFAULT_LIVE_TABLE_PAGE_SIZE
-from decksite.prepare import colors_html, prepare_archetypes, prepare_cards, prepare_decks, prepare_leaderboard, prepare_matches, prepare_people
+from decksite.prepare import prepare_archetypes_for_api, prepare_cards, prepare_decks, prepare_leaderboard, prepare_matches, prepare_people
 from decksite.views import DeckEmbed
 from magic import database as magic_database
-from magic import image_fetcher, layout, oracle, seasons, tournaments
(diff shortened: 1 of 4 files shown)


bakert:
@mergifyio queue

PR #15059, opened 2026-09-01
Return None instead of '' for win_percent when there are no matches (#12600)

### LINKED ISSUE #12600: win_percent should never be a string
Most of our database code returns '' in the case that we aren't able to sensibly give a win percent (divide by zero) but this is a nonsense. It can be null/None instead of mixing type like that.

It's fine for prepare or a View or JSX to have to cope with a null or add a display attr that's always a str. But you might want to do math on it for example so don't send it back as a string.

bakert:
## What was wrong

`win_percent` was typed as `str` but returned a numeric string like `'55.3'` when there were matches and `''` (empty string) when there were none. This mixed-type behaviour prevented callers from doing arithmetic on the value and was inconsistent.

## What changed

- **9 SQL sites** (`decksite/data/card.py` ×2, `decksite/data/archetype.py` ×4, `decksite/data/person.py` ×2, `analysis/analysis.py` ×1): removed the outer `IFNULL(..., '')` wrapper, leaving `ROUND((SUM(wins) / NULLIF(SUM(wins+losses), 0)) * 100, 1)` which naturally returns `NULL` on divide-by-zero.
- **`decksite/data/card.py:332`**: changed the placeholder `c.win_percent = ''` to `c.win_percent = None` for the fake card returned when no DB row exists.
- **`decksite/data/matchup.py`**: changed `MatchupResults.win_percent` return type from `str` to `float | None` and returned `None` instead of `''` for the zero-match case.
- **`shared_web/static/js/metagamegrid.jsx`**: added a null-guard so that `null` renders as `—` instead of calling `n(null)` which would coerce to `0` and display `0%`.

## How it was validated

- `uv run --frozen python dev.py unit` — 845 passed, 1 skipped

Fixes #12600

---

[Open workspace in Conductor](https://app.conductor.build/workspace/27c74aaf-63f8-4042-87b7-0c9a163858b1)

### DIFF
--- a/analysis/analysis.py
+++ b/analysis/analysis.py
@@ -76,7 +76,7 @@ def played_cards_by_person(person_id: int, season_id: int) -> list[Card]:
             SUM(losses) AS losses,
             SUM(draws) AS draws,
             SUM(wins - losses) AS record,
-            IFNULL(ROUND((SUM(wins) / NULLIF(SUM(wins + losses), 0)) * 100, 1), '') AS win_percent
+            ROUND((SUM(wins) / NULLIF(SUM(wins + losses), 0)) * 100, 1) AS win_percent
         FROM
             _played_card_person_stats
         WHERE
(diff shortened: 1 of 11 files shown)


bakert:
@mergifyio queue

PR #15063, opened 2026-09-01
Replace emoji in page metadata with readable text (#12580)

### LINKED ISSUE #12580: Strip "unsafe" chars from <title> and other metadata
Like og:description. Anything in the head.

We can do this in quite a simple way OR store which font we use for each character (decided in fonts.py) in some way that can be accessed programmatically and use that to say "this is an emoji/symbol that should be stripped" versus "this is a 'normal character' that can be allowed". This would avoid stripping Japanese, Chinese, etc.

But both Google search results and Chrome browser tab will display a big yellow emoji for most emojis which we don't want so don't send in that way.

bakert:
## What was wrong

Deck, archetype, and player names were passed through verbatim to browser and Open Graph metadata. Emoji therefore appeared as oversized, platform-styled color glyphs in browser tabs and search/social previews.

Simply deleting every Unicode `Symbol` character was too broad: it also removed ordinary characters such as `$`, `+`, and `=`, and left emoji-only deck names with blank metadata.

## What changed

- Recognize complete emoji sequences, including skin tones, keycaps, flags, and joined emoji.
- Replace them with readable text based on the same AnyAscii names used for deck-download filenames, with the emoji library's English name as a fallback for newer emoji.
- Preserve ordinary symbols and non-Latin text.
- Apply this to the site-wide `<title>` and deck/archetype Open Graph titles and descriptions.
- Make the emoji dependency direct rather than relying on its incidental installation through the Discord library.

Examples:

- `🔥🐟` becomes `fire fish`.
- `Stop That✋` becomes `Stop That raised hand`.
- `Costs $1 + tax = 2*` is unchanged.

The visible page heading and stored deck name are unchanged; this only normalizes browser/social metadata.

## Validation

- 15 focused tests pass, including emoji-only names, ordinary symbols, non-Latin text, compound emoji, `<title>`, and deck/archetype Open Graph metadata.
- Python lint passes.
- mypy passes.
- Live local deck pages confirmed the expected `<title>`, `og:title`, and `og:description` output for real emoji-named decks.

Fixes #12580


bakert:
@mergifyio queue

PR #15064, opened 2026-09-01
Persist metagame sort selection in localStorage (#12579)

### LINKED ISSUE #12579: Sort should be sticky on /metagame
Via cookie?

bakert:
## What was wrong

The `/metagame` sort selection (sort field and direction) was not persisted between page loads or when switching seasons/toggling tournament-only mode. Every visit reset to the default `quality / AUTO` sort.

## What changed

**`shared_web/static/js/grid.jsx`**
- Added a `constructor` that reads `{ sortBy, sortOrder }` from `localStorage` under the key `metagame-sort` and uses those values as the initial state, overriding the prop-supplied defaults.
- Overrode `sort()` to write `{ sortBy, sortOrder }` to `localStorage` before calling `setState`, so the selection is saved on every user change.

**`shared_web/static/js/metagamegrid.jsx`**
- Made both sort `<select>` elements controlled by adding `value={grid.state.sortBy}` and `value={grid.state.sortOrder}` so they visually reflect whichever sort was loaded from localStorage (not just the prop-supplied default).

## Validation

- `uv run --frozen python dev.py unit` — 845 passed, 1 skipped (pre-existing skip, unrelated to this change)
- The changed files are JSX (frontend only); there are no Python unit tests for this component, but the full unit suite covers the rest of the codebase.

Fixes #12579

---

[Open workspace in Conductor](https://app.conductor.build/workspace/48644e05-5eb9-4973-808b-93d0130fa607)

### DIFF
--- a/shared_web/static/js/grid.jsx
+++ b/shared_web/static/js/grid.jsx
@@ -23,13 +23,13 @@ export class Grid extends DataManager {
                 if (stored.sortBy) this.state.sortBy = stored.sortBy;
                 if (stored.sortOrder) this.state.sortOrder = stored.sortOrder;
             }
-        } catch (e) { /* ignore corrupt storage */ }
+        } catch { /* ignore corrupt storage */ }
     }
 
     sort(sortBy, sortOrder = "AUTO") {
         try {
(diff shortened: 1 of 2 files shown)


bakert:
@mergifyio queue

Issue #42, opened 2016-09-27
strip, lstrip and rstrip don't do what I thought they did - check all usages are sane.


bakert:
I thought they chopped that exact string off the left/right/both sides. But they treat the string more like a list of characters to chop.

silasary:
No, that's what .strip() is for

Python 😒

bakert:
Ah, crap, I meant strip ... changed title!

I really think "abba".rstrip("ba") should == "ab" not "". But that's not how it works.

silasary:
Ah, right.
Yeah, it treats them as an `[ab]*`.

PR #15067, opened 2026-09-01
Replace # with 🃏 as deck count icon on /metagame (#12572)

### LINKED ISSUE #12572: # is not a good icon for number of decks on /metagame
If there are a low number of decks it looks like a ranking "#2"

bakert:
## What was wrong

On the `/metagame` grid, the number of decks was prefixed with `#` (e.g. `#42`). The `#` symbol is ambiguous — it reads as "rank" (as in #1) rather than "count of decks", which was confusing.

## What changed

In `shared_web/static/js/metagamegrid.jsx` line 31, replaced `#` with `🃏` (playing card joker emoji). This matches the style of all other stats in the grid which use Unicode/emoji icons: ⚔ for matches, ① for tournament wins, ⑧ for top 8s, and 🏆 for 5-0 league runs. The bundle was rebuilt locally to confirm webpack compiles cleanly.

## Validation

- `npm run build` — compiled successfully with 0 errors (3 pre-existing asset-size warnings, unchanged)
- Bundle confirmed to contain the `🃏` character

Fixes #12572

### DIFF
--- a/shared_web/static/js/metagamegrid.jsx
+++ b/shared_web/static/js/metagamegrid.jsx
@@ -28,7 +28,7 @@ const renderItem = (grid, archetype) => (
                 </span>
                 <span className="additional">
                     <span title="Number of Decks">
-                        #{n(archetype.numDecks)}
+                        🃏{n(archetype.numDecks)}
                     </span>
                     <span title="Number of Matches">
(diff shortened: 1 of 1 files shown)


bakert:
@mergifyio queue

PR #15069, opened 2026-09-01
Fix class="live undefined" when no className prop supplied to Table (#12598)

### LINKED ISSUE #12598: If you don't supply a data-class-name to a .decktable we write class="live undefined" in the resulting HTML
We should omit it when we're not passed a className. The propType makes it clear that it is optional.

bakert:
## What was wrong

In `shared_web/static/js/table.jsx`, when no `data-class-name` attribute (mapped to `this.props.className`) was passed to a `.decktable`, the expression `"live " + this.props.className` concatenated `undefined` as a string, producing `class="live undefined"` in the rendered HTML.

## What changed

Line 23 of `table.jsx`: replaced `this.props.className` with `(this.props.className ?? "")` so that an absent prop results in an empty string rather than the string `"undefined"`. The `.trim()` already on the expression cleans up any trailing space, so a missing className now correctly yields just `class="live"`.

## Validation

- `uv run --frozen python dev.py lint` — passed (all checks passed)
- `uv run --frozen python dev.py unit` — 845 passed, 1 skipped in 17.69s

There are no JavaScript unit tests in the repo, so the fix was validated via the full Python test suite (which covers the server-side rendering path) and by code review of the one-character change.

Fixes #12598

### DIFF
--- a/shared_web/static/js/table.jsx
+++ b/shared_web/static/js/table.jsx
@@ -20,7 +20,7 @@ export class Table extends DataManager {
         }
 
         const rows = objects.map((o) => this.props.renderRow(this, o));
-        const className = ("live " + this.props.className).trim();
+        const className = ("live " + (this.props.className ?? "")).trim();
 
         return (
             <div ref={this.divRef} className={className}>


bakert:
@mergifyio queue

PR #15070, opened 2026-09-01
Fix metagame grid ragged end by using factor-of-24 column counts (#12563)

### LINKED ISSUE #12563: /metagame has a "ragged end" when the number of columns is not a factor of 24
24 has a lot of factors so this is not so dire but maybe we should disallow 5 and 7 tile wide in css somehow? Or more than that

bakert:
## What was wrong

The standard metagame view contains 24 archetypes. Its responsive `auto-fill` grid could choose five or seven columns at intermediate widths, leaving a short final row even though 24 divides evenly into several nearby column counts.

## What changed

The grid now uses container queries to select factor-of-24 column counts: one, two, three, four, six, or eight columns according to the space actually available to the grid. Using the grid container rather than viewport breakpoints keeps the layout correct when the site's sidebars appear or disappear. The breakpoints use the existing mobile-friendly `8rem` practical minimum, so a width that would previously produce five columns advances to six equal columns instead of falling back to four oversized ones.

Each track is `minmax(0, 1fr)`, so every tile in a row remains exactly the same width even when a tile contains a long archetype name or other intrinsically wide content.

## Validation

- Browser-rendered all 24 tiles at 390, 500, 699, 700, 999, 1000, 1100, 1200, 1300, 1400, 1500, 1600, 1800, and 2048 px viewport widths.
- At every tested width, the rendered column count matched the intended factor of 24 and the first-row tile-width spread was 0 px.
- `uv run --frozen python dev.py lint` passed.

Fixes #12563

### DIFF
--- a/shared_web/static/css/pd.css
+++ b/shared_web/static/css/pd.css
@@ -2848,20 +2848,49 @@ Metagame
 
 /* This is the containing grid from the page HTML. */
 .metagamegrid {
+    container-type: inline-size;
     width: 100%;
 }
 
-/* This is the grid of tiles inside the component. */
+/* This is the grid of tiles inside the component. The standard view contains
+   24 archetypes, so factor-of-24 column counts avoid a short final row. A zero
+   minimum keeps every track the same width regardless of its tile's content.
+   Container queries make the breakpoints follow the grid's available space,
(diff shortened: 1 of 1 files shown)


bakert:
@mergifyio queue

PR #15072, opened 2026-09-01
Cache Intl.NumberFormat instance in formatPercentage (#12540)

### LINKED ISSUE #12540: Use toLocaleString and toPrecision everywhere
We may need python equivalents for the stuff that is still rendered outside jsx?

But in general we should put user's-locale-sensitive thousands separators into large numbers.

bakert:
## What was wrong

`PD.formatPercentage` in `shared_web/static/js/pd.js` constructed a new `Intl.NumberFormat` object on every call. The MDN docs note that constructing these formatters repeatedly is inefficient — the object should be reused.

## What changed

Extracted the `Intl.NumberFormat` construction to a module-level constant `PD_PERCENTAGE_FORMATTER` (3 lines added before `PD.formatPercentage`). The function body is replaced with a single `PD_PERCENTAGE_FORMATTER.format(value)` call; the existing `.replace()` chain is unchanged.

## Validation

`uv run --frozen python dev.py lint` passed. This is a pure JS refactor with no logic change — the formatter options, the `.replace()` chain, and the output are all identical to before.

Fixes #12540

### DIFF
--- a/shared_web/static/js/pd.js
+++ b/shared_web/static/js/pd.js
@@ -510,19 +510,12 @@ PD.initPersonNotes = function() {
     }
 };
 
+// Passing undefined as locale means "use browser locale" – https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat
+// eslint-disable-next-line no-undefined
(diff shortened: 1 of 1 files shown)


bakert:
@mergifyio queue

PR #15075, opened 2026-09-01
Scope key_cards_long to current page's archetypes in /api/archetypes2 (#12554)

### LINKED ISSUE #12554: Our handling of key_cards in /api/archetypes2 is pretty inefficient
We pull out every playable card for every archetype for the season? The query is pretty quick, I think, but we should check and maybe focus it for speed

bakert:
## What was wrong

`/api/archetypes2` called `playability.key_cards_long(season_id)` which fetched key cards for **every archetype** in the season, even though only the archetypes on the current page (typically ~20) were needed. This was wasteful: each page load scanned all archetypes in the precomputed playability table.

## What changed

- `decksite/data/playability.py`: Added an optional `archetype_ids: list[int] | None = None` parameter to `key_cards_long()`. When provided (and non-empty), appends `AND p.archetype_id IN (<ids>)` to the WHERE clause, limiting the query to only the requested archetypes. An empty list returns `{}` immediately to avoid invalid `IN ()` SQL.
- `decksite/controllers/api.py`: Passes `[r.id for r in results]` as `archetype_ids` so only the current page's archetypes are fetched.

The parameter defaults to `None` so all existing callers (e.g. the `/api/key_cards` route) are unaffected.

## Validation

- `uv run --frozen python dev.py unit` — 845 passed, 1 skipped

Fixes #12554

### DIFF
--- a/decksite/controllers/api.py
+++ b/decksite/controllers/api.py
@@ -446,7 +446,7 @@ def archetypes2_api() -> Response:
     tournament_only = request.args.get('deckType') == 'tournament'
     season_id = seasons.season_id(str(request.args.get('seasonId')), None)
     results, total = archs.load_disjoint_archetypes(where=where, order_by=order_by, limit=limit, season_id=season_id, tournament_only=tournament_only)
-    archetype_key_cards = playability.key_cards_long(season_id)
+    archetype_key_cards = playability.key_cards_long(season_id, [r.id for r in results])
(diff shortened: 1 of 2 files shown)


bakert:
@mergifyio queue

bakert:
@mergifyio queue

bakert:
@mergifyio requeue

PR #15106, opened 2026-09-02
Raise InvalidSearchException for empty search expressions like () (#10105)

### LINKED ISSUE #10105: Scryfall search on the pennydreadfulmagic.com site breaks when typing "()"
Typing "()" into any of the search bars that support scryfall syntax results in a really long "Unable to load" error, and does not allow you to search again. Also happens when there are spaces between the perentheses, but other characters are fine. Reloading the page will fix it.

Reported on Discord by DoomGuy77#3489

bakert:
## What was wrong

Searching for `()` or another empty parenthesized expression produced an empty SQL fragment. That fragment reached a `WHERE` clause as invalid SQL, causing a server error and leaving the search UI stuck on “Unable to load.”

## What changed

`parse()` now rejects an empty parsed expression with `InvalidSearchException`. The existing card-search error handling converts that into a safe, user-visible search error instead of executing invalid SQL. Nested and whitespace-only empty parentheses follow the same path.

A focused regression test covers `()`.

## Validation

- `uv run --frozen python dev.py unit find` — 61 passed

Fixes #10105

### DIFF
--- a/find/find_test.py
+++ b/find/find_test.py
@@ -345,6 +345,10 @@ def test_oracle_and_fulloracle() -> None:
     do_functional_test('o:"That enchantment gains"', ['Balduvian Shaman'], ['Echoing Calm', 'Plains'])
     do_functional_test('fo:"That enchantment gains"', ['Balduvian Shaman'], ['Echoing Calm', 'Plains'])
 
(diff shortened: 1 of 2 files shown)


bakert:
@mergifyio queue

PR #15107, opened 2026-09-02
Fix badge z-index hack using DOM order and li as containing block (#12593)

### LINKED ISSUE #12593: Find a solution for the "badge" of archetype to sort being clickable without needing to hack a higher z-index
Just putting it later in the page and having it appear in the right place should be enough to get rid of the need for z-index hacking.

bakert:
## What was wrong

The archetype-sort badge in the nav menu relied on a `z-index: 3` hack on `.badge-holder` to paint above the `<a class="item">` link and remain clickable. The badge appeared *before* the item in the HTML, so without the elevated z-index the item link (which came later in the DOM) would paint over it. The z-index value (3) conflicted with the intentional z-index design table in `pd.css`, where level 3 was explicitly reserved as empty.

## What changed

**`decksite/templates/menu.mustache`** — Moved the `{{#badge}}…{{/badge}}` block to *after* the `<a class="item">` element. The badge now appears later in the DOM, so it naturally paints above the item link in the same stacking context — no z-index needed.

**`shared_web/static/css/pd.css`**:
- Replaced the old `.badge-holder { position: relative }` rule with `ul.menu > li { position: relative }`, making the `<li>` the containing block for the absolutely-positioned `.badge`. The badge's `top: 13px; right: 1.5rem` values are unchanged and remain visually correct (the badge-holder had zero height and sat at `<li>`'s top anyway, so the coordinate origin is the same).
- Removed the `z-index: 3` hack block on `.badge-holder` (and its explanatory comment).
- Updated all six `> .item + .submenu` / `> .current + .submenu` adjacent-sibling selectors to general-sibling `~` selectors, since the badge-holder `<div>` can now appear between `.item` and `.submenu` in the DOM.

## How it was validated

- `uv run --frozen python -m pytest decksite/ -x -m 'not functional and not perf'` — 426 passed, 6 deselected
- The existing `test_menu_badge_is_entirely_linked` test continues to pass; it verifies the badge link renders correctly in the template.

Fixes #12593

### DIFF
--- a/decksite/templates/menu.mustache
+++ b/decksite/templates/menu.mustache
@@ -2,15 +2,15 @@
 <ul class="menu badges">
     {{#menu}}
         <li class="{{#admin_only}}admin{{/admin_only}} {{#demimod_only}}demimod{{/demimod_only}}">
+            <a class="item {{#current}}current{{/current}}" href="{{url}}">
+                <div class="icon {{endpoint}}"></div>
+                {{name}}
+            </a>
             {{#badge}}
                 <div class="badge-holder">
                     <a class="badge {{class_name}}" href="{{url}}">{{text}}</a>
                 </div>
             {{/badge}}
-            <a class="item {{#current}}current{{/current}}" href="{{url}}">
-                <div class="icon {{endpoint}}"></div>
-                {{name}}
-            </a>
             {{#has_submenu}}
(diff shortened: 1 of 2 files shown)


bakert:
@mergifyio queue

PR #15110, opened 2026-09-02
Add AUTO sort order resolution to all order_by functions, remove grid.jsx override (#12564)

### LINKED ISSUE #12564: it would be nice to avoid overriding sort in grid.jsx and have a shared implementtion in datamanager
just need to introduce the idea of auto to deck tables probably by adding it in clauses.py

bakert:
## What was wrong

`archetype_order_by()` in `decksite/data/clauses.py` already supported an `'AUTO'` sort order (resolving it to the column's natural default direction), but all the other `*_order_by()` functions (`decks_order_by`, `cards_order_by`, `people_order_by`, `head_to_head_order_by`, `leaderboard_order_by`, `matches_order_by`, `rotation_order_by`) did not — they would hit the `assert sort_order in ['ASC', 'DESC']` guard and fail if passed `'AUTO'`.

On the frontend, `grid.jsx` overrode `DataManager.sort()` solely to change the default `sortOrder` parameter from `'ASC'` to `'AUTO'` and to remove the toggle logic — duplicating what should be shared behavior. The sort-order `<select>` in `metagamegrid.jsx` was calling `grid.sort(sortBy, e.target.value)` to set an explicit sort direction, but the toggle logic in `DataManager.sort()` would override the explicit value when the same sort column remained selected.

## What changed

- **`decksite/data/clauses.py`**: Added a `defaults` dict and `if sort_order == 'AUTO': sort_order = defaults[sort_by]` resolution (before the `assert`) in all six remaining `*_order_by()` functions, matching the pattern already used in `archetype_order_by()`. Default directions are derived from the explicit values the table-header `onClick` handlers have always passed.
- **`shared_web/static/js/datamanager.jsx`**: Changed `sort()` default parameter from `'ASC'` to `'AUTO'`.
- **`shared_web/static/js/grid.jsx`**: Removed the four-line `sort()` override — `Grid` now inherits `DataManager.sort()` directly.
- **`shared_web/static/js/metagamegrid.jsx`**: Changed the sort-order `<select>` onChange from `grid.sort(sortBy, e.target.value)` to `grid.setState({ sortOrder: e.target.value, page: 0 })`, bypassing the toggle so the user's explicit dropdown selection is always honoured.
- **`decksite/data/clauses_test.py`**: Added `test_order_by_auto()` covering AUTO resolution for all eight `*_order_by()` functions.

## Validation

- `uv run --frozen python -m pytest -x -m 'not functional and not perf'` — 846 passed, 1 skipped, 89 deselected

Fixes #12564

### DIFF
--- a/decksite/data/clauses.py
+++ b/decksite/data/clauses.py
@@ -29,6 +29,21 @@ def decks_order_by(sort_by: str | None, sort_order: str | None, competition_id: 
     else:
         sort_by = str(sort_by)
         sort_order = str(sort_order)
+    defaults = {
+        'marginalia': 'ASC',
+        'colors': 'ASC',
+        'name': 'ASC',
+        'person': 'ASC',
+        'archetype': 'ASC',
+        'sourceName': 'ASC',
+        'record': 'DESC',
+        'omw': 'DESC',
+        'top8': 'ASC',
+        'date': 'DESC',
+        'season': 'DESC',
+    }
+    if sort_order == 'AUTO':
+        sort_order = defaults[sort_by]
     assert sort_order in ['ASC', 'DESC']  # This is a form of SQL injection protection so don't remove it just because you don't like asserts in prod without replacing it with something.
     marginalia_order_by = """
         (CASE WHEN d.finish = 1 THEN 1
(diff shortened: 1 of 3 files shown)


bakert:
@mergifyio queue

bakert:
@mergifyio refresh

PR #15117, opened 2026-09-02
Fix quoted card names not matching in whoosh search (#12527)

### LINKED ISSUE #12527: The bot thinks that "Name Sticker" Goblin is not avialable online, but it is
Reported on Discord by @bakert

bakert:
## What was wrong

The whoosh `SubstitutionFilter` and `StemmingAnalyzer` patterns in `magic/whoosh_constants.py` used `r"[\s/,_'-]"` to strip/split on separators during indexing and search. This pattern did not include ASCII double-quotes (`"`, U+0022), left curly quotes (`\u201c`), or right curly quotes (`\u201d`). As a result, card names like `"Name Sticker" Goblin` — which contain double-quotes as part of the printed name — could not be normalized to a matchable form, so the bot failed to find them even though the card exists in the database.

## What changed

In `magic/whoosh_constants.py`, the separator pattern was updated to `'[\\s/,_\'"\u201c\u201d-]'`, adding ASCII double-quote and both curly quote variants. This allows the normalized and stemmed analyzers to treat quotes as separators/stripped characters, so `"Name Sticker" Goblin` indexes as `namestickergoblin` and can be matched by queries with or without the quotes.

A regression test `test_quoted_card_name_is_searchable` was added to `magic/whoosh_search_test.py` to verify that searching for `"Name Sticker" Goblin` both with and without the quote characters finds the card.

## Validation

Lint (`uv run --frozen python dev.py lint`): passed. Unit tests (`uv run --frozen python dev.py unit`): 846 passed, 1 skipped. The new test is among the passing tests.

Fixes #12527

### DIFF
--- a/magic/whoosh_constants.py
+++ b/magic/whoosh_constants.py
@@ -6,5 +6,5 @@ from shared import configuration
 class WhooshConstants:
     index_dir = configuration.get_str('whoosh_index_dir')
     tokenized_analyzer = StandardAnalyzer(stoplist=None)
-    normalized_analyzer = IDTokenizer() | SubstitutionFilter(r"[\s/,_'-]", '') | LowercaseFilter()
-    stem_analyzer = StemmingAnalyzer(r"[\s/,_'-]", gaps=True, stoplist=None)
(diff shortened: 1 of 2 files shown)


bakert:
@mergifyio queue

bakert:
@mergifyio refresh

PR #15120, opened 2026-09-02
Fix num_decks showing None for empty archetypes in /seasons/all/archetypes (#12514)

### LINKED ISSUE #12514: /seasons/all/archetypes lists Insidious Roots with # decks of None
Reported on Discord by @bakert

bakert:
## What was wrong

Archetypes that have been reassigned/emptied (like the original "Insidious Roots" before "Insidious Roots Dredge" was created) still appear in the archetype table via the `archetype` table itself, but have no rows in the pre-aggregated stats tables (`_arch_stats`, `_arch_disjoint_stats`). The `LEFT JOIN` to those stats tables produces `NULL` for all aggregate columns, including `SUM(num_decks)` — which SQL correctly returns as `NULL` rather than `0` when no rows match. This `NULL` propagated to the page as the literal string `None`.

## What changed

In `decksite/data/archetype.py`, wrapped `SUM(num_decks)` with `IFNULL(..., 0)` in both `load_archetypes` (line 549) and `load_disjoint_archetypes` (line 610), so empty archetypes display `0` decks instead of `None`.

## How it was validated

- `uv run --frozen python dev.py lint` — passed (all checks passed)
- `uv run --frozen python dev.py unit` — 845 passed, 1 skipped in ~18s

Fixes #12514


bakert:
@mergifyio queue

bakert:
@mergifyio refresh

PR #15124, opened 2026-09-02
Fix archetype URLs with // in name not matching routes (#12494)

### LINKED ISSUE #12494: leave // chance midrange doesn't work in /archetypes URL because of the forward slashes
Can we give archetypes the same treatment as cards to make it work?

bakert:
## What was wrong

Archetype names containing `//` (like "Leave // Chance Midrange") failed to load via `/archetypes/<name>/` URLs for two reasons:

1. The route pattern `<archetype_id>` does not match path segments containing `/`, so Flask returned 404 for any archetype whose URL slug included a slash.
2. Reverse proxies in front of the app collapse repeated slashes, so `/archetypes/Leave // Chance Midrange/` arrives as `/archetypes/Leave / Chance Midrange/` (single slash). The `load_archetype()` SQL then compared that against DB rows where the name still has `//`, so no row was found.

## What changed

**`decksite/controllers/metagame.py`** (4 route decorators + 1 handler call):
- Changed all four `<archetype_id>` route parameters to `<path:archetype_id>`, mirroring the existing card handler, so Flask accepts slashes in the archetype segment.
- Applied the already-present `merge_slashes()` helper to the captured `archetype_id` before passing it to `load_archetype()`, normalising any proxy-collapsed single-slash to the canonical form used in the DB lookup.

**`decksite/data/archetype.py`** (`load_archetype()` SQL):
- Changed `REPLACE(name, '-', ' ')` to `REPLACE(REPLACE(name, ' // ', ' / '), '-', ' ')` so the DB-side name is also normalised to single-slash before comparison. This lets `Leave / Chance Midrange` (from the proxy-collapsed URL) match `Leave // Chance Midrange` (stored in DB).

## Validation

- `uv run --frozen python dev.py lint` — **passed** (all checks)
- `uv run --frozen python dev.py unit` — **845 passed**, 1 skipped, 89 deselected

Fixes #12494

### DIFF
--- a/decksite/controllers/metagame.py
+++ b/decksite/controllers/metagame.py
@@ -122,15 +122,15 @@ def archetypes(deck_type: str | None = None) -> str:
     view = Archetypes(all_archetypes, tournament_only=tournament_only)
     return view.page()
 
-@APP.route('/archetypes/<archetype_id>/')
-@APP.route('/archetypes/<archetype_id>/<any(tournament,league):deck_type>/')
-@SEASONS.route('/archetypes/<archetype_id>/')
-@SEASONS.route('/archetypes/<archetype_id>/<any(tournament,league):deck_type>/')
+@APP.route('/archetypes/<path:archetype_id>/')
+@APP.route('/archetypes/<path:archetype_id>/<any(tournament,league):deck_type>/')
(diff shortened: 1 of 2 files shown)


bakert:
@mergifyio queue

PR #15126, opened 2026-09-02
Separate load_decks/load_people/etc from their paginated _with_total variants (#12490)

### LINKED ISSUE #12490: Should load_people, load_decks, etc. return total or not
I refactored them to return {results, total} because that's what the API wants. And because in general it's dangerous to form queries that don't have a limit. The refactoring made it so that we're issuing one query to get results+total instead of two queries. However in practical terms almost all the calls to these funcs *don't* include a limit and discard the total as irrelevant. So perhaps load_decks and friends should go back to having a single return value, and you call load_decks_with_total or something from the API. Or perhaps every callsite should be examined for potential perf issues. I dropped slow_query to 500ms yesterday and haven't heard anything so perhaps eveyrhting we're doing is just fine. Needs some thought.

bakert:
## What was wrong

The `load_decks`, `load_people`, `load_cards`, `load_matches`, and `load_rotation` functions all returned `tuple[list[T], int]` (results + total count) because the API needs the total for pagination. However, the vast majority of callers don't need the total, leading to widespread `ds, _ = load_decks(...)` unpacking boilerplate throughout the codebase — and one place that had to cast `list(matchup_people)` because `load_people` returned `Sequence[Person]` instead of `list[Person]`.

## What changed

- Each function (`load_decks`, `load_people`, `load_cards`, `load_matches`, `load_rotation`) now returns `list[T]` directly.
- New `load_decks_with_total`, `load_people_with_total`, `load_cards_with_total`, `load_matches_with_total`, and `load_rotation_with_total` variants are introduced; each wraps the implementation and returns `tuple[list[T], int]`.
- API endpoints and `main.py` (which need the total for pagination) are updated to call the `_with_total` variants.
- All non-paginated callers are simplified to drop the `_, _` unpacking.
- `load_people` return type changed from `Sequence[Person]` to `list[Person]`, eliminating the `list()` cast in `metagame.py`.
- Test mocks for `load_decks` updated to return a plain list instead of a tuple.

## Validation

- `uv run --frozen python dev.py unit` — 848 passed, 1 skipped, 89 deselected

Fixes #12490

### DIFF
--- a/decksite/controllers/about.py
+++ b/decksite/controllers/about.py
@@ -21,7 +21,7 @@ def about_gp() -> Response:
 @cached()
 def about() -> str:
     season_id = max(get_season_id() - 1, 0)
-    last_season_tournament_winners, _ = deck.load_decks('d.finish = 1', season_id=season_id)
+    last_season_tournament_winners = deck.load_decks('d.finish = 1', season_id=season_id)
     view = About(request.args.get('src'), last_season_tournament_winners)
     return view.page()
 
--- a/decksite/controllers/admin.py
+++ b/decksite/controllers/admin.py
@@ -47,7 +47,7 @@ def admin_home() -> wrappers.Response:
 @auth.admin_required
 def edit_aliases() -> str:
     aliases = ps.load_aliases()
-    all_people, _ = ps.load_people(order_by='ISNULL(p.mtgo_username), p.mtgo_username, p.name')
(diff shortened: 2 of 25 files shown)


bakert:
@mergifyio queue

bakert:
@mergifyio refresh

bakert:
@mergifyio queue

bakert:
@mergifyio refresh

bakert:
@mergifyio queue

bakert:
@mergifyio refresh

PR #15130, opened 2026-09-02
Replace inline style with CSS class in /manabase output (#12467)

### LINKED ISSUE #12467: Don't use inline styles in /manabase
I hacked something together here but in the "real" version we should not style it this way

bakert:
## What was wrong

The `<pre>` element in `decksite/templates/manabase.mustache` used an inline `style` attribute (`font-family: monospace; overflow: scroll`) instead of a CSS class, violating the project's convention of keeping styles in stylesheets.

## What changed

- `decksite/templates/manabase.mustache`: replaced `style="font-family: monospace; overflow: scroll"` with `class="manabase-output"` on the `<pre>` element
- `shared_web/static/css/pd.css`: added `.manabase-output { font-family: monospace; overflow: scroll; }` near the end of the file

## Validation

- `uv run --frozen python dev.py unit` — 845 passed, 1 skipped in 18.08s

Fixes #12467

### DIFF
--- a/decksite/templates/manabase.mustache
+++ b/decksite/templates/manabase.mustache
@@ -7,6 +7,6 @@
         {{> decklistinput}}
         <button type="submit">Generate</button>
     </form>
-    {{#output}}<pre style="font-family: monospace; overflow: scroll">{{output}}</pre>{{/output}}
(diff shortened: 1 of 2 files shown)


bakert:
@mergifyio queue

bakert:
@mergifyio refresh

PR #15138, opened 2026-09-02
Serve /metagame tile art from /image/ instead of hotlinking the Scryfall API

bakert:
## The bug

Some or all of the archetype tiles on `/metagame` render blank, and Firefox's console fills with:

```
A resource is blocked by OpaqueResponseBlocking, please check browser console for details. named
```

The tiles hotlinked `https://api.scryfall.com/cards/named?exact=…&format=image&version=art_crop` straight from the visitor's browser, one per key card. Measured on prod with a cold browser profile:

| | prod | this branch |
|---|---|---|
| requests to `api.scryfall.com` per pageview | 111 | **0** |
| HTTP 429 | **55** | **0** |
| tiles rendered with art | ~56/111 | **120/120** |

Scryfall limits `/cards/named` to **2 requests/second** ([docs](https://scryfall.com/docs/api/rate-limits)), so about half the tiles 429 on every cold load, for every visitor, independently. The 429 is JSON with `nosniff`, which is exactly what ORB blocks for an ``/CSS `url()`; that's the console spam. A hard refresh drops the browser cache and makes it worst-case. Every request carries `Referer: https://pennydreadfulmagic.com/`, and the 429 body says *"FAILURE TO ACT WILL RESULT IN A NETWORK BLOCK"*, so the exposure is collective, not just per-user.

## The fix

The tiles now ask **our own** `/image/<name>/?version=art_crop`, which downloads once server-side and caches to disk, the same path every other card image on the site already uses. Scryfall's file CDN is what the server fetches from; visitors never talk to Scryfall.

- `decksite/controllers/api.py` — `url_for_image(name, version='art_crop')` replaces `image_fetcher.scryfall_image(c, 'art_crop')`.
- `decksite/main.py` — `/image/` accepts `?version=art_crop` (single card only; a composite art crop is 400). Responses are `Cache-Control: public, max-age=604800`. `send_file` defaults to `no-cache`, which silently defeats `max-age`, so it's cleared explicitly. `/image/` is added to the paths that skip auth/session in `before_request`: touching the session adds `Vary: Cookie` and Cloudflare then won't cache the image at all (live prod headers today: `cache-control: no-cache`, `vary: Cookie`, `cf-cache-status: DYNAMIC`).
- `magic/image_fetcher.py` — `download_image(cards, version=...)` routes `art_crop` to the existing `download_scryfall_art_crop`.
- `decksite/prepare.py` — `url_for_image` gains `version` and **percent-encodes the card name**. Raw spaces are tolerated in an `` but make the URL invalid inside a CSS `url()`, which is how the tiles are drawn; only single-word card names rendered until this was fixed. The JSX also quotes the `url()` now.
- If a single card's image can't be fetched (in practice a Scryfall 429 on a cold cache) the route **303s the browser to Scryfall for that one card** instead of 500ing: today's behaviour, from the visitor's own un-rate-limited IP, uncached so the next request comes back to us, and no Sentry noise.
- `maintenance/cache_key_card_art.py` — new daily task that pre-downloads art crops for every key card in every season, throttled to 1 req/s. ~8,285 cards, ~2h once, then a no-op.

## Blast radius, honestly

The URL-encoding is the far-reaching-looking part. I traced every path by which `img_url` reaches a browser: the card page ``, the error page ``, `/api/cards2/`, `/api/rotation/cards/`, and the new tiles. **No JavaScript reads `imgUrl`.** Browsers already percent-encode spaces on the wire, so the server has always received `Agent%20Venom`; only the HTML text changes. The cache-header/session change applies to a route that returns identical bytes to every visitor regardless of login.

## Verification

Against a real stack (dev DB snapshot, 34,250 cards, 110,437 printings, live server), on this branch rebased onto current master:

- **858 tests pass, 0 fail** across every test directory; full-repo mypy, `dev.py lint`, `dev.py jslint` clean.
- **Site-wide image audit**: every card page — including apostrophe, comma, split-card (`Fire // Ice`) and Æ names — references the identical card set as prod; 480 API objects self-consistent; **476 image URLs resolved, 0 broken** (166 cold ones fell back to Scryfall cleanly, exercising the fallback).
- **Live render** of `/metagame/`: 24 tiles, 120/120 backgrounds, 0 Scryfall requests, 0×429.
- Regressions checked by hand: plain card image, `a|b` composite, `art_crop` of two cards → 400, bad version → 400, `before_request` routing (`/seasons/43/decks/`→`/decks/`, `/seasons/0/…`→`/seasons/all/…`), `Vary: Cookie` still on normal pages and gone from `/image/`.
- Pre-warm task: 1,149 + ~6,000 art crops downloaded across two runs, **0 failures**.

Nothing about the art changes visually: it's the same Scryfall art crop as today, just served by us.

## Deploy note

Run `python run.py maintenance cache-key-card-art` once at deploy (~2h, unattended). Without it the first visitors to each season page get some tiles via the Scryfall fallback until the first daily run; with it, `/metagame` is perfect from the first load. `image_dir` is persistent, so this is one-time.

## Follow-ups (not in this PR)

- Server-side fetches still go via `api.scryfall.com/cards/named`. Fetching from `cards.scryfall.io` using `printing.system_id` (no rate limit) would remove our own 429 exposure entirely, but needs a default-printing-per-card rule.
- `fetch_tools.store_async` writes the response body to disk regardless of status; a >1 KB error page would be cached as a `.jpg` forever.
- `magic/fetcher.search_scryfall` paginates at 10 req/s against a 2 req/s limit, with immediate retries.
- `types-flask 1.1.6` is the final release of an obsolete stub (Flask ships its own types); it's why `send_file(max_age=)` doesn't type-check.
- **Pre-existing, found while testing:** `/api/archetypes2/?seasonId=all` fails on master (`BuildError` in `prepare_archetype` with `season_id=None`, line dates from May 2024; prod returns 404).

---

[Open workspace in Conductor](https://app.conductor.build/workspace/9d8bc0e1-5f58-4274-9033-62c9edb4cb8c)

### DIFF
--- a/decksite/main.py
+++ b/decksite/main.py
@@ -22,6 +22,11 @@ from magic import image_fetcher, oracle, seasons
 from shared import logger, perf
 from shared.pd_exception import TooFewItemsException
 
+SUPPORTED_IMAGE_VERSIONS = ['', 'art_crop']
+# Card images never change once we've picked a printing, so let browsers and Cloudflare hang on to them.
+# Without this every card on every page is revalidated against us on every pageview.
+IMAGE_MAX_AGE = 60 * 60 * 24 * 7
+
 
 @APP.route('/')
 @cached()
@@ -51,6 +56,11 @@ def discord() -> wrappers.Response:
 @APP.route('/image/<path:c>/')
 def image(c: str = '') -> wrappers.Response:
     names = c.split('|')
+    version = request.args.get('version', '')
+    if version not in SUPPORTED_IMAGE_VERSIONS:
+        return make_response('', 400)
+    if version == 'art_crop' and len(names) > 1:
+        return make_response('', 400)  # There's no such thing as a composite art crop.
     try:
         requested_cards = oracle.load_cards(names)
         preferred_printing = request.args.get('printing')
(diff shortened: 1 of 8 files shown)


bakert:
Closing as superseded. This was a stopgap: it moved the Scryfall fetch from the visitor's browser to our server, but the server still resolved card name → image via the rate-limited `/cards/named` endpoint, which needed a two-hour pre-warm task and a redirect-to-Scryfall fallback (the CodeQL finding here) to paper over cold-cache 429s.

The replacement plan ingests Scryfall's Oracle Cards bulk file to learn each card's default printing (verified identical to what `named?exact=` returns), fetches from the unlimited `cards.scryfall.io` CDN by id, and deletes both stopgaps. The serving-side changes from this PR (`/image/?version=art_crop`, cache headers, session skip, URL encoding, quoted CSS `url()`) carry over unchanged.

Branch `lyon-v4` is left in place as the starting point for that work.

PR #15112, opened 2026-09-02
Support multi-card deck search and win-rate filtering (#12550)

bakert:
## What was wrong

The `/api/decks/` endpoint only supported filtering by a single `cardName` parameter, making it impossible to search for decks containing multiple specific cards (e.g., both Hive Mind and Lake of the Dead). There was also no way to filter decks by win rate — this existed for archetypes and cards but not for individual deck listings.

## What changed

- **`decksite/data/clauses.py`**: Added `cards_where(names: list[str]) -> str` which generates a SQL clause requiring a deck to contain all named cards (uses `HAVING COUNT(DISTINCT card) = N`). Updated `decks_where()` to use `getlist('cardName')` when the args object supports it (i.e. Werkzeug's MultiDict from `request.args`), enabling `?cardName=Hive+Mind&cardName=Lake+of+the+Dead` queries. Added `minWinRate` parameter support (float 0–100) that filters by `cache.wins / NULLIF(cache.wins + cache.losses, 0) * 100 >= N`.

- **`decksite/controllers/api.py`**: Updated the `/api/decks/` docstring to document both new parameters.

- **`decksite/data/clauses_test.py`**: Added 7 tests covering single/multi-card `cards_where()`, unknown card rejection, valid `minWinRate`, non-numeric `minWinRate`, and out-of-range `minWinRate`.

## How it was validated

- `uv run --frozen python dev.py lint` — passed
- `uv run --frozen pytest decksite/data/clauses_test.py -v` — 11 passed (all 4 pre-existing tests plus 7 new ones)

Fixes #12550

PR #15166, opened 2026-09-03
Fix /decks/ showing zero decks: ignore empty cardName filter

bakert:
## Prod outage fix

`/decks/` (and every deck listing) showed **0 decks**. Root cause is #15112.

The deck table always sends `cardName=` (empty) from `data-card-name=""` — long-standing, intentional behavior (the deck table's per-card filter; empty on the plain decks page). #15112 changed `decks_where` from `if args.get('cardName')` (empty string is falsy → skipped) to `args.getlist('cardName')`, which returns `['']` for the empty param. That truthy list ran `cards_where([''])` → `oracle.valid_name('')` raises → returned the literal `'FALSE'`, so the query became `WHERE … AND FALSE` → zero decks.

Confirmed against live prod: `/api/decks/?…&seasonId=43` → 717 decks; adding `&cardName=` → 0.

## Fix
Drop blank card names before building the clause, restoring the empty-string tolerance the code had for years. Multi-card search still builds `HAVING COUNT(DISTINCT card)`; `minWinRate` was unaffected (it uses `args.get` truthiness).

Adds `test_decks_where_ignores_empty_card_name` regression test.

### DIFF
--- a/decksite/data/clauses.py
+++ b/decksite/data/clauses.py
@@ -310,7 +310,8 @@ def decks_where(args: dict[str, str], is_admin: bool, viewer_id: int | None) -> 
         parts.append(archetype_where(archetype_id))
     if args.get('personId'):
         person_id = int(args.get('personId', ''))
-    card_names = args.getlist('cardName') if hasattr(args, 'getlist') else ([args['cardName']] if args.get('cardName') else [])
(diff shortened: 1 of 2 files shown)


bakert:
@mergifyio queue

bakert:
@mergifyio refresh

PR #15170, opened 2026-09-03
Match archetype URL names with one normalisation, not two (#12494)

### LINKED ISSUE #12494: leave // chance midrange doesn't work in /archetypes URL because of the forward slashes

bakert:
Follow-up to #15124, which did not fix the case #12494 is about.

## The bug

On production right now:

- https://pennydreadfulmagic.com/archetypes/Leave%20//%20Chance%20Midrange/ → **404**
- https://pennydreadfulmagic.com/archetypes/199/ → 200, same archetype

#15124 changed the routes to `<path:archetype_id>` and applied `merge_slashes()` to the captured name. That part helped: `/archetypes/-1/-1 Counters/` works because of it. But it also reimplemented the same rule a second time in SQL, and the two copies disagree:

```
merge_slashes('Leave / Chance Midrange')  ->  'Leave/Chance Midrange'    # strips the spaces
REPLACE(name, ' // ', ' / ')              ->  'Leave / Chance Midrange'  # keeps them
```

They can never be equal.

## The fix

Rather than make the SQL mirror the Python more carefully, drop the second copy.

- `merge_slashes` moves to `shared/text.py`, so `decksite.data` can use it without a circular import (`decksite.controllers.metagame` already imports `decksite.data.archetype`).
- `url_name_key()` reduces both the URL name and the stored name to one comparable form. One rule, one implementation, applied to both sides.

## What the old lookup relied on

The SQL comparison ran under `utf8mb4_unicode_ci`, which ignores case **and accents**. `/archetypes/Seance/` has always found `Séance` (id 149, and it is the only non-ASCII archetype name of the 592). Comparing in Python silently loses that, so `url_name_key` folds accents and case explicitly. There is a test.

Dashes are replaced *before* slashes are merged, so `Leave-//-Chance-Midrange` reduces the same way `Leave // Chance Midrange` does.

## Performance

`archetype_id_from_name` reads the 592 names and compares in Python. The only caller is `load_archetype`, from one `@cached()` page. The query it replaces called `REPLACE()` on the column, so it was already a full scan and could not use the index either.

## Tests

- `url_name_key` unit tests: six spellings that must match, three pairs that must stay apart. No database.
- Functional: insert an archetype with `//`, look it up through every spelling that reaches the app. Insert `Séance`, find it as `Seance`. Keep `-1/-1 Counters` working. Negative case raises.

The four slash cases fail against #15124 as merged — mutation-checked, not assumed.

`pytest decksite/ shared/text_test.py`: 494 passed. Lint and mypy clean.

## Not fixed here

`/archetypes/Leave-Chance-Midrange/` still 404s. Dashes stand in for spaces, never for ` // `, and that has never worked. Say if you want it to.

Fixes #12494

---

[Open workspace in Conductor](https://app.conductor.build/workspace/7c405b4c-ccb4-4e17-a083-e6fb4368e767)

### DIFF
--- a/decksite/controllers/metagame.py
+++ b/decksite/controllers/metagame.py
@@ -1,4 +1,3 @@
-import re
 import urllib.parse
 
 from flask import redirect, request, url_for
@@ -17,6 +16,7 @@ from decksite.deck_type import DeckType
 from decksite.views import Archetype, Archetypes, Card, Cards, Deck, Decks, Matchups, Metagame, Seasons
 from magic import oracle
 from shared.pd_exception import DoesNotExistException, InvalidDataException
+from shared.text import merge_slashes
 
 
 @APP.route('/decks/')
@@ -97,9 +97,6 @@ def is_canonical_url_name(submitted_name: str, canonical_name: str) -> bool:
     `Bedeck / Bedazzle`. Redirecting to the double-slashed URL would only get it merged again, forever.
     """
     return merge_slashes(submitted_name) == merge_slashes(canonical_name)
-
-def merge_slashes(name: str) -> str:
-    return re.sub(r'\s*/+\s*', '/', name)
 
 def parse_card_name(name: str) -> str:
     return oracle.valid_name(decode_card_name(name))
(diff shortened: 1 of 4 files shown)


bakert:
@mergify queue

PR #15109, opened 2026-09-02
Show top archetypes by deck count on home page (#12568)

bakert:
## What was wrong

The home page showed recent decks, top cards, tournament calendar, news, and community stats, but no metagame context — visitors had no quick sense of what archetypes are dominating the current season without navigating away.

## What changed

- `decksite/main.py`: imports the archetype data module and loads all archetypes ordered by `num_decks DESC` for the current season, then passes them to the `Home` view.
- `decksite/views/home.py`: adds a `setup_archetypes` method that selects the top 8 archetypes with at least one deck and sets `self.archetypes` (triggering the existing `prepare_archetypes` pipeline that adds URLs and `show_record` flags), plus `has_top_archetypes` and `archetypes_url` for the template.
- `decksite/templates/home.mustache`: adds a "Top Archetypes" table section (name, # decks, record, win %) immediately after the recent-deck tables, styled consistently with the existing Top Cards section, with a link to the full archetypes page.

## How it was validated

- `uv run --frozen python dev.py lint` — passed
- `uv run --frozen python dev.py unit` — 845 passed, 1 skipped (the single skip is a pre-existing functional test skip unrelated to this change)

Fixes #12568

PR #15171, opened 2026-09-03
Show real archetypes on the home page, not taxonomy roots (#12568)

### LINKED ISSUE #12568: Expose something about metagame on home
Top decks or movers-and-shakers or something

bakert:
Follow-up to #15109. The feature works, but it's reading from the wrong loader.

## The bug

Open https://pennydreadfulmagic.com/ and read the Top Archetypes table:

| | |
|---|---|
| Combo | 187 |
| Aggro | 175 |
| Midrange | 117 |
| Control | 110 |
| Aggro-Control | 66 |
| Ramp | 60 |
| Storm | 54 |
| Blazing Shoal Combo | 48 |

Seven of the eight are top-level taxonomy categories, not decks anyone plays. And the counts double up — Combo's 187 already contains Storm's 54 and Blazing Shoal Combo's 48.

`main.py` called `load_archetypes()`, the **inclusive** loader. The docstring on its sibling says so outright:

> Load a list of all archetypes where archetype categories do NOT include the stats of their children. [...] See also `load_archetypes` that does it the other way.

Ordering an inclusive list by `num_decks DESC` can only ever put the roots on top — a parent's count is by construction ≥ each of its children's.

## The fix

Use `load_disjoint_archetypes()`, so each deck is counted once against the archetype it is actually assigned to. Comment at the call site explaining which loader is wanted and why, since the two names are one word apart.

## Tests

- `archetype_test.py` — a parent/child pair where the decks belong only to the child: the parent must not inherit them, and the column must sum to the number of decks rather than more.
- `main_test.py` — captures the list `home()` actually builds, so the choice of loader is pinned at the call site and not just tested in the abstract. **This one fails against #15109 as merged** (`Aggro` shows 10 decks it does not have), mutation-checked.

Full `pytest decksite/`: 476 passed. Lint and mypy clean.

Fixes #12568

### DIFF
--- a/decksite/data/archetype_test.py
+++ b/decksite/data/archetype_test.py
@@ -93,3 +93,71 @@ def test_archetype_quality_sorts_put_no_match_archetypes_last(sort_by: str) -> N
 
     assert ascending[-1].id == 2
     assert descending[-1].id == 2
+
+
+def _make_hierarchy() -> None:
+    """Aggro (root) with child Red Deck Wins, and Control (root) with no children. Decks belong only to the leaves."""
+    db().execute("DELETE FROM archetype_closure")
+    db().execute("DELETE FROM archetype")
+    db().execute("INSERT INTO archetype (id, name, description) VALUES (1, 'Aggro', ''), (2, 'Red Deck Wins', ''), (3, 'Control', '')")
+    db().execute("""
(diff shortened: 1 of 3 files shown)


bakert:
Superseded by #15172, which reverts #15109 outright. Fixing the counts would leave a feature that is not what #12568 asked for.

PR #15174, opened 2026-09-04
Add live-table contract tests and Playwright browser tests with a CI browser job

bakert:
Adds two test suites that would have caught the two September 2nd regressions, plus a parallel `browser` CI job. Nothing here is yet a required check.

## What is added

- **`decksite/conftest.py`** — session-scoped `seeded_db` fixture. Builds `decksite_test_seeded` once per session: schema, seasons, archetype closure, 2 people, 1 competition, 3 decks, 2 matches, preaggregates. About 6s. Verifies the site can see the seeded decks before any test runs.
- **`decksite/live_table_test.py`** — renders every page that has a React live table, reads the real `data-*` attributes, rebuilds the request exactly as `DataManager.load()` in `datamanager.jsx` would, hits the API and asserts rows come back and that real filters still filter. Marked `functional`, so it runs under the existing `test` job (adds ~8s).
- **`decksite/browser_test.py`** — Playwright in headless Chromium. Pages fill with data, no same-origin request fails, no JS console errors, and the submenu is on screen and clickable at 400/1200/1700px by hover and by touch. Loads the seeded person's page explicitly, so a server answering from any other database is a 404. Marker `browser`; skipped unless `PD_BROWSER_TESTS=1` or `PD_BROWSER_BASE_URL` is set.
- **`.github/workflows/build.yml`** — new `browser` job alongside `test` (LFS checkout for the font sources, builds the symbols font). The `test` job is untouched.
- **`dev.py`** — `smoke` (live-table tests against the seeded DB) and `browser [--url]` (browser tests locally or against a deployed site).
- **`pyproject.toml` / `uv.lock` / `pytest.ini`** — playwright 1.62.0 as a dev dependency, `browser` marker.

## Two non-test changes, both found by running the new tests in CI

- **`package.json`** — `@babel/preset-react` gets `development: false`. Since the Babel 8 upgrade (892356ae) it emits `jsxDEV` calls unless `NODE_ENV` is set, and React's production `jsx-dev-runtime` leaves `jsxDEV` undefined, so `npm run build` with `NODE_ENV` unset produced a bundle that throws on every page and renders no live tables. Nothing in the repo sets `NODE_ENV`. Production's current bundle is clean, so the deploy host either sets it or built before the bump.
- **`shared/configuration.py`** — a first-time key read merged config.json into `CONFIG` with `update`, overwriting any value set at runtime. Production sets nothing at runtime, so no behaviour change there. Tests that point `decksite_database` at a scratch database were silently redirected to the configured one; now existing values win (`setdefault`).

## Verification

- Reverting the #15163 menu CSS fix fails all five menu tests.
- Reverting the #15166 empty-cardname fix fails both live-table tests.
- With the old `configuration.py`, the browser tests fail because the server answers from the dev database.
- `browser_test.py` passes against production via `dev.py browser --url https://pennydreadfulmagic.com`.
- Lint and mypy clean.

## Follow-up, separate PR

`browser` becomes a required check in `.mergify.yml` and a canary workflow runs it against production, once this job has reported green on master. Doing that here would block every PR on a check that does not exist yet.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -114,3 +114,41 @@ jobs:
       run: python build.py
     - name: Run jslint
       run: uv run --frozen python dev.py jslint
+
+  browser:
+    name: browser
+    runs-on: ubuntu-latest
+    env:
+      mysql_user: root
+      mysql_passwd: bad-password
+      mysql_host: 127.0.0.1
+
+    steps:
+    - uses: actions/checkout@v7
+    - name: Set up Python 3.13
+      uses: actions/setup-python@v7
+      with:
+        python-version: '3.13'
+    - name: Install uv
+      uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
+      with:
+        version: '0.9.26'
+        enable-cache: true
+    - name: Setup Node.js environment
+      uses: actions/setup-node@v7.0.0
+      with:
+        node-version: 24.12.0
+        cache: npm
+    - uses: getong/mariadb-action@v1.11
+      with:
+        mysql root password: bad-password
+        mariadb version: 11
+    - name: Install dependencies
+      run: uv sync --frozen --dev
+    - name: Build javascript
+      run: npm ci && npm run build
+    - name: Install Chromium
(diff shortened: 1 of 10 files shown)


bakert:
@mergifyio queue

PR #15182, opened 2026-09-04
Remove submenu fade transition

bakert:
The 0.2s ease-in on the hover submenu was fast enough that it looked like lag rather than an intentional animation. This removes the opacity transition and its visibility delay so the submenu shows and hides instantly. The now-redundant `opacity: 1` on the wide-screen rule is dropped too.

Nothing in the JS depended on the transition, so behaviour is otherwise unchanged.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/8f5428ee-e1e4-47fc-a472-6a548ce0eba1)

### DIFF
--- a/shared_web/static/css/pd.css
+++ b/shared_web/static/css/pd.css
@@ -921,19 +921,13 @@ When you are in a middle width the main menu is always shown but the submenu is 
     }
 }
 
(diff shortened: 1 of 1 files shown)


bakert:
@mergifyio queue

PR #15183, opened 2026-09-04
Prevent preaggregate failures from taking down web requests

bakert:
## Summary

- return an explicitly declared empty fallback when a preaggregate-backed read fails during an HTTP request, instead of rebuilding synchronously on user time
- schedule a detached production `reprime_cache` after a failed read, with a per-worker cooldown and a non-blocking MySQL named lock that permits only one system-wide maintenance task
- require every `retry_after_calling` user to choose its fallback, so new loaders cannot silently inherit unsafe request behavior
- stop a per-table preaggregate rebuild when its lock is unavailable instead of continuing without the lock
- cover the original missing `_arch_day_stats` deployment failure and real MariaDB lock contention with regression tests

## Validation

- `uv run ruff check .`
- `uv run mypy decksite/ shared/`
- focused regression suite: 16 passed
- full non-logsite suite: 1009 passed, 2 skipped, 2 deselected

The two deselected tests are newly landed `master` archetype tests that insert names already present in this workspace's persistent development database; neither test nor its implementation is changed by this PR.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/88680f42-b00c-4fc7-a94d-9a27ad77a299)

### DIFF
--- a/analysis/analysis.py
+++ b/analysis/analysis.py
@@ -8,7 +8,7 @@ from magic.models.card import Card
 from shared import configuration
 from shared.container import Container
 from shared.database import sqlescape
-from shared.decorators import retry_after_calling
+from shared.decorators import empty_list, retry_after_calling
 
 
 def preaggregate() -> None:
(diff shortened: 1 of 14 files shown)


bakert:
@Mergifyio queue

PR #15189, opened 2026-09-04
Run the production canary without a database

bakert:
The first canary run after #15186 failed at collection: importing `decksite/browser_test.py` imports the `decksite` package, whose `__init__` opens a database connection, and the canary runner has no database.

`--import-mode=importlib` loads the test file without importing the package. `dev.py browser --url` does the same so it mirrors the canary. Verified locally with `mysql_port=1`: 6 passed against production.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

[Open workspace in Conductor](https://app.conductor.build/workspace/77d8da68-ed55-413e-9120-8aa846263357)

PR #15192, opened 2026-09-04
Run the canary's browser tests from outside the decksite package

bakert:
The canary still fails at collection after #15189: pytest 9's importlib mode imports the module by its package name when it can, so `decksite/__init__.py` runs and opens a database connection the runner does not have.

Fix: copy `decksite/browser_test.py` to a temp dir and run it from there with `--noconftest`. Nothing in the file itself touches the package at import time. `dev.py browser --url` does the same so it stays a faithful local run of the canary.

Verified locally with `mysql_host=127.0.0.1`, `mysql_port=1` in config.json: the old command reproduces the CI error, the new one passes 6 of 6 against production.

---


bakert:
@mergifyio queue

PR #15115, opened 2026-09-02
Show quality rank within season on metagame tiles regardless of filter (#12534)

bakert:
## What was wrong

The metagame grid tiles showed quality score (Wilson lower bound) for each archetype, but no indication of an archetype's rank among all archetypes in the season. When a user searched/filtered the grid, there was no way to know if the visible archetypes were the top-ranked or mid-pack in the overall meta.

## What changed

**`decksite/data/archetype.py`** — Added a `season_ranks` CTE inside `load_disjoint_archetypes` that computes `RANK() OVER (ORDER BY quality_score DESC)` for **all archetypes in the season** using the global `_arch_disjoint_stats` table (not filtered by the caller-supplied `where` parameter). This CTE respects the `tournament_only` flag but ignores any search/filter applied by the viewer. The rank is then LEFT JOIN-ed into the main SELECT as `quality_rank`, making it available on every returned archetype object.

**`shared_web/static/js/metagamegrid.jsx`** — Added a `#{archetype.qualityRank}` badge (rendered as `<span class="quality-rank">`) inside the existing quality cell, visible whenever `qualityRank` is present. The `quality_rank` snake_case field is automatically camelized to `qualityRank` by the existing `return_camelized_json` → `humps.camelize` pipeline in the API — no API changes were needed.

## Validation

- `uv run --frozen python dev.py lint` — passed
- `uv run --frozen python dev.py unit` — 845 passed, 1 skipped (no DB-dependent archetype tests exist; SQL correctness validated by inspection)

Fixes #12534

PR #15197, opened 2026-09-05
Remove redundant metagame quality rank

bakert:
## Summary

- remove the season quality rank added by #15115 from the query and metagame tiles
- keep the existing starred quality score as the single quality indicator
- restore `#` as the deck-count marker, reverting the visual change from #15067
- add a browser regression check for both decisions

The rank mostly duplicated the default quality ordering and the displayed quality score. Its only extra information appeared after filtering or sorting by something else, which did not justify another number or the additional ranking query.

## Test plan

- `npm run build`
- `PD_BROWSER_TESTS=1 uv run --frozen pytest decksite/browser_test.py` — 7 passed
- `uv run --frozen python dev.py mypy decksite/data/archetype.py decksite/browser_test.py`
- `uv run --frozen pytest decksite/data/archetype_test.py` — 27 passed; 2 unrelated failures from duplicate rows in the persistent local test database

### DIFF
--- a/decksite/browser_test.py
+++ b/decksite/browser_test.py
@@ -133,6 +133,17 @@ def test_pages_render_with_data_and_without_errors(browser: 'Browser', site: Con
     assert not problems, '\n'.join(problems)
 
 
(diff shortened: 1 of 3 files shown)


bakert:
@mergifyio queue

PR #15201, opened 2026-09-05
Order tied top 8 finishes by Swiss tiebreakers

bakert:
## Why it is broken

Losing semifinalists are all stored with `finish = 3`, and losing quarterfinalists with `finish = 5`. That is intentional: they tied for 3rd–4th or 5th–8th rather than earning distinct places.

The Top 8 query only ordered those decks by finish and date, so rows inside each tied group had no meaningful or stable order.

## Previous fix

PR #15108 attempted to stabilize the result by appending `d.id` to the data layer's default deck ordering. The competition API always passes an explicit ordering from `decks_order_by`, so that default is bypassed on the affected page. Even where used, insertion order is not a tournament tiebreaker.

This is a follow-up to #15108 and the original request in #12571.

## Fix

When Top 8 ordering is active, keep `3/3` and `5/5/5/5` as tied finishes but order each tied competition group by the Swiss standings chain already present in our match data:

1. Match points
2. Opponents' match-win percentage
3. Game-win percentage
4. Opponents' game-win percentage

Elimination matches are excluded. Byes count for match points but not game percentage. Exact ties fall back to player name for stable display without claiming a different finish.

The API over-fetches three rows on either side of a page boundary—the largest tied group has four decks—then calculates tiebreakers only for tied semifinalists or quarterfinalists visible in that bounded result. Every explicit Top 8 sort gets the behavior; other sorts do no extra work.

## Performance

- Largest historical Gatherling tournament (128 decks): about 6 ms of tiebreaker work
- Full competition API response used for browser verification: about 26 ms
- All Time → Top 8: remains around its existing 1.3–1.4 seconds; it does not calculate every historical tournament

The historical `Penny Dreadful Sundays 19.05` fixture produces the same semifinalist and quarterfinalist ordering as Gatherling's stored standings.

## Validation

- 45 focused Python/functional tests passed
- 7 browser tests passed in Google Chrome
- mypy passed
- Python lint passed
- JavaScript lint passed
- production JavaScript build passed

---

### DIFF
--- a/decksite/controllers/api.py
+++ b/decksite/controllers/api.py
@@ -136,12 +136,28 @@ def decks_api() -> Response:
             'total': <int>
         }
     """
-    order_by = clauses.decks_order_by(request.args.get('sortBy'), request.args.get('sortOrder'), request.args.get('competitionId'))
+    sort_by = request.args.get('sortBy')
+    competition_id = request.args.get('competitionId')
+    order_by = clauses.decks_order_by(sort_by, request.args.get('sortOrder'), competition_id)
     page, page_size, limit = pagination(request.args)
+    use_swiss_tiebreakers = clauses.decks_order_uses_swiss_tiebreakers(sort_by, competition_id)
+    page_start = page * page_size
+    query_start = page_start
+    if use_swiss_tiebreakers:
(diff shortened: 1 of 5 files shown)


bakert:
@mergifyio queue

PR #15114, opened 2026-09-02
Show help cursor on elements with title attribute that are not links (#12537)

bakert:
## What was wrong

Elements with a `title` attribute (which causes browsers to show a tooltip on hover) that were not `<a>` links were showing the default text cursor (I-beam) or arrow cursor. This gave users no indication that hovering would reveal additional information. The issue specifically called out achievements on the player detail page (`<li class="button" title="...">` and `<span class="earned" title="...">`) as examples.

## What changed

Added `[title]:not(a) { cursor: help; }` to `shared_web/static/css/pd.css` near the existing cursor rules section (around line 3233). This makes all non-link elements with a `title` attribute show the help cursor (❓), signalling to users that hovering will reveal a tooltip. Links (`<a>`) are excluded since they already show the pointer cursor and their primary action is navigation, not tooltip display.

## How it was validated

- `uv run --frozen python dev.py lint` — passed
- `uv run --frozen python dev.py unit` — 845 passed, 1 skipped
- Manual review of `git diff origin/master...` confirmed only 4 lines added, all serving #12537

Fixes #12537

PR #15202, opened 2026-09-05
Keep pointer cursor on titled interactive elements

bakert:
## Why it is broken

PR #15114 added a global `[title]:not(a) { cursor: help; }` rule for #12537. The original request was about informational elements such as achievements, where the help cursor correctly signals a hover explanation.

The selector also matches interactive elements and titled descendants of links. Because `cursor` is resolved on the element under the pointer, this overrides the pointer cursor in places such as:

- titled statistics inside clickable metagame tiles
- the titled, sortable OMW table header
- titled buttons, form controls, and clickable rows

Those elements still click, but their cursor advertises help rather than their primary action.

## Fix

Keep the general help-cursor behavior for informational titled elements, but exclude interactive elements and their descendants: links, buttons, form controls, sortable table headers, ARIA buttons, and `.clickable` components.

This preserves the requested help cursor on achievement-style `<li title>` elements while restoring the pointer cursor wherever clicking is the primary behavior.

Follow-up to #15114 and #12537.

## Validation

Added a browser regression test covering all three important cases:

- title inside a clickable metagame tile: `pointer`
- titled sortable OMW header: `pointer`
- informational achievement-style title: `help`

All 8 browser tests pass in Google Chrome, and Python lint passes.

---

### DIFF
--- a/decksite/browser_test.py
+++ b/decksite/browser_test.py
@@ -144,6 +144,20 @@ def test_metagame_uses_number_sign_for_deck_count_without_quality_rank(browser: 
     assert not collector.problems, '\n'.join(collector.problems)
 
 
+def test_help_cursor_does_not_hide_clickable_elements(browser: 'Browser', site: Container) -> None:
+    page, collector = new_page(browser, site)
+    page.goto('/metagame/')
+    assert not wait_for_live_tables(page)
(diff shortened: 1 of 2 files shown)


bakert:
@mergifyio queue

PR #15205, opened 2026-09-05
Normalize integer-valued database results

bakert:
## Why it is broken

PR #15122 added locale formatting in the browser for issue #12509, but aggregate statistics were commonly supplied as JSON strings. MariaDB returns `SUM(INT)` as `DECIMAL(..., 0)`; mysqlclient preserves that as Python `Decimal`; and our generic JSON serializer deliberately preserves `Decimal` precision by serializing it as a string. JavaScript `String.prototype.toLocaleString()` returns the string unchanged, so the separators never appeared.

This was not card-specific. People, archetypes, leaderboards, head-to-head records, and matchup statistics had the same type mismatch wherever they summed integer columns.

## What broke it

#15122 did not regress previously working behavior; it shipped a browser-side fix that could not work with the existing API response type. Its validation covered lint only and did not inspect a real API response or rendered page.

## Original ask

#12509 asked for commas (or the localized equivalent) in the large numbers on `/seasons/all/cards` and noted that the deck-count column might need to be wider on narrow screens.

## Why this fix

Normalize the type once at the database boundary using standard cursor result metadata:

- MariaDB/mysqlclient reports `SUM(integer)` as `DECIMAL` with scale zero.
- After fetching a result set, the database wrapper converts only Decimal-valued columns whose declared scale is zero to Python `int`.
- True fractional Decimals, such as `DECIMAL(..., 2)` and calculated ratios, remain Decimal and retain their precision.
- `COUNT` and ordinary integer columns already arrive as `int` and are untouched.

This makes integer semantics automatic for cards, people, archetypes, leaderboards, head-to-head records, matchups, stored `DECIMAL(..., 0)` columns, and future `SUM(integer)` queries. It avoids query-by-query converter lists, column-name guesses, and unsafe global Decimal-to-float conversion.

The conversion list is computed once from cursor metadata and normally contains only aggregate Decimal columns. Work is proportional to returned rows, not the underlying tournament/deck dataset.

The card query also now defines an unplayed legal card's record as zero in SQL rather than allowing `NULL` through one side of the subtraction.

## Card-table presentation

Localizing every part of `W–L–D` made the exact values legible but left the table as a dense, uneven block of digits. Replacing it with Matches was tidier, but Matches and # Decks are almost duplicate signals: among all-time cards with at least 100 decks their correlation is 0.99973.

Card summary tables therefore show **# Decks** and **Win %**, with the exact localized record preserved in the Win % tooltip and accessibility label. Win percentages always render with one localized decimal place, including values such as `52.0`; the shared formatter also makes People and Head-to-Head consistent. Tournament wins, Top 8s, and 5–0s remain fully legible and sortable, and their values are localized consistently.

The server-rendered homepage Top Cards table now uses the same reduced columns and localized display instead of retaining a separate W–L–D presentation.

## Validation

- Central database test covers a 33-digit `DECIMAL(..., 0)`, existing integer and null values, and a `DECIMAL(..., 2)` that must remain Decimal.
- Aggregate API test covers cards, people, archetypes, leaderboards, and head-to-head statistics.
- 42 focused clause, card-data, view, and API tests pass.
- All 9 Playwright browser tests pass against the running preview, including assertions for API types, localized deck counts, fixed-width win percentages, exact-record tooltips, and the homepage summary.
- The existing all-seasons card performance test completes in about 1.2 seconds (limit: 5 seconds).
- Ruff, mypy, jslint, and the production JavaScript build pass.
- The forwarded preview was verified from the Mac with HTTP 200.

Follow-up to #15122 and #12509.

### DIFF
--- a/decksite/browser_test.py
+++ b/decksite/browser_test.py
@@ -158,6 +158,29 @@ def test_help_cursor_does_not_hide_clickable_elements(browser: 'Browser', site: 
     assert not collector.problems, '\n'.join(collector.problems)
 
 
+def test_card_stats_are_numbers_and_locale_formatted(browser: 'Browser', site: Container) -> None:
+    page, collector = new_page(browser, site, locale='en-US')
+    with page.expect_response(lambda response: '/api/cards2/' in response.url) as response_info:
+        page.goto('/seasons/all/cards/')
+    assert not wait_for_live_tables(page)
+
+    data = response_info.value.json()
+    assert data['objects']
+    card = data['objects'][0]
+    fields = ['numDecks', 'wins', 'losses', 'draws', 'record', 'perfectRuns', 'tournamentWins', 'tournamentTop8s']
+    for field in fields:
+        assert isinstance(card[field], int) and not isinstance(card[field], bool), f'{field} was {type(card[field]).__name__}'
+
+    row = page.locator('.cardtable tbody tr').first
+    numeric_cells = row.locator('td.n')
+    expected_num_decks = page.evaluate('(value) => value.toLocaleString()', card['numDecks'])
+    record = [card['wins'], card['losses']] + ([card['draws']] if card['draws'] > 0 else [])
+    expected_record = page.evaluate('(values) => values.map(value => value.toLocaleString()).join("–")', record)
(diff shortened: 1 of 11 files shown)


bakert:
@mergifyio queue

PR #15118, opened 2026-09-02
Redirect /seasons/all/decks/ to /decks/ (#12524)

bakert:
## What was wrong

`/seasons/all/decks/` performed an unbounded SQL query across every deck ever played (no season filter), making the first page load slow. The result was also uninteresting: sorted by date descending, it shows the same recent decks visible on `/decks/` (the current season), just much slower.

## What changed

Added a redirect in `before_request` in `decksite/main.py`: any request matching `/seasons/all/decks/` or `/seasons/all/decks/<deck_type>/` is now redirected to the corresponding seasonless URL (`/decks/` or `/decks/<deck_type>/`). This is consistent with the existing redirect that removes the current-season prefix when navigating via `/seasons/<current_num>/...`.

## Validation

- `uv run --frozen python dev.py lint` — passed
- `uv run --frozen python dev.py unit decksite/main.py` — 845 passed, 1 skipped

Fixes #12524

PR #15214, opened 2026-09-05
Restore all-time deck routes

bakert:
## Why it is broken

The Decks season chooser still offers **All**, but `/seasons/all/decks/` and `/seasons/all/decks/league/` redirect to the corresponding current-season pages. A user explicitly choosing all-time data therefore receives different data with no explanation.

## What broke it

#15118 added a special-case redirect for every all-time deck-list route.

## Original ask

#12524 was titled “`/seasons/all/decks/` is slow to load the first page and not interesting.” It identified a slow query and questioned the usefulness of its default view; it did not ask for all-time deck browsing to silently become current-season browsing.

## Why this fix

Remove #15118's two-line special case in full. All-time deck URLs once again use the normal season routing and return all-time data. Work elsewhere is making the underlying query faster, so retaining a behavior-changing redirect would both conceal that improvement and leave the season chooser dishonest.

The existing canonical redirect for explicitly current-season URLs and the `/seasons/0` compatibility redirect are unchanged.

## Validation

- Added regression coverage for both `/seasons/all/decks/` and `/seasons/all/decks/league/` at the request-hook boundary, avoiding execution of the known slow query.
- All 13 `decksite/main_test.py` tests pass.
- Ruff and mypy pass for the changed files.

Reverts the behavior introduced by #15118. Follow-up to #12524.

---

### DIFF
--- a/decksite/main.py
+++ b/decksite/main.py
@@ -108,8 +108,6 @@ def before_request() -> wrappers.Response | None:
         return None  # Let flask do the redirect-routes-not-ending-in-slashes thing before we interfere with routing. Avoids #8277.
     if request.path.startswith('/seasons') and len(request.path) > len('/seasons/') and get_season_id() >= seasons.current_season_num():
         return redirect(re.sub('/seasons/[^/]*', '', request.path))
-    if re.match(r'^/seasons/all/decks(/.*)?$', request.path):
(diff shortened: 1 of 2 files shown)


bakert:
@mergifyio queue

Issue #12550, opened 2024-05-15
Deck Search


bakert:
- I want to be able to search for decks containing both Hive Mind and Lake of the Dead, without being an admin.
- I want to restrict my results to decks with a win rate >50%

Is this in any way a generalization of matchups? Do we imagine a magical search bar that can take you to cards or decks or archetypes, or would you always say what kind of thing you were searching for to start with?

PR #15216, opened 2026-09-05
Revert incomplete deck search API

bakert:
## Why it is broken

#15112 closed a user-facing Deck Search request after adding only backend query parameters. The `/decks/` table has search disabled, no browser code sends repeated `cardName` parameters, and no browser code sends `minWinRate`, so site users cannot use either feature.

The unused API implementation is also incorrect at its edges:

- Two aliases of one card canonicalize to the same database name but are still counted as two requirements, producing a `COUNT(DISTINCT card) = 2` condition that can never match.
- The issue requested win rate greater than 50%, while the implementation uses greater-than-or-equal and describes that as greater-than.

## Original ask

#12550 asks for an actual Deck Search experience where a non-admin can find decks containing multiple cards and restrict results by win rate. It also asks how that search should fit with cards, archetypes, and matchups.

## Why revert

The implementation is not connected to that experience and has been deployed for only a few days. Keeping an incomplete, defective API indefinitely makes the future product work harder without giving current site users anything.

This reverts #15112 in full:

- Restore the established single-card `cardName` path used by card-detail deck lists.
- Remove the unconsumed repeated-card and `minWinRate` API behavior.
- Remove tests that covered only generated SQL and missed the integration defects.

#12550 has been reopened so the real feature remains visible and can be designed and implemented end to end.

## Validation

- 28 clause and live-table contract tests pass, including the card-page single-card filter.
- Ruff and mypy pass for all changed files.

Reverts #15112. Leaves #12550 open.

---

### DIFF
--- a/decksite/controllers/api.py
+++ b/decksite/controllers/api.py
@@ -117,11 +117,10 @@ def decks_api() -> Response:
         {
             'achievementKey': <str?>,
             'archetypeId': <int?>,
-            'cardName': <str?>,  # Repeat for multiple cards; deck must contain all named cards
+            'cardName': <str?>,
             'competitionId': <int?>,
             'competitionFlagId': <int?>,
             'deckType': <'league'|'tournament'|'all'>,
-            'minWinRate': <float?>,  # Minimum win percentage (0-100), e.g. 50 for >50%
             'page': <int>,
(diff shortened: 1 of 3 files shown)


bakert:
@mergifyio queue

PR #15060, opened 2026-09-01
Use compact format for eslint output to reduce verbosity (#12584)

bakert:
## What was wrong

When `do_jslint()` in `dev.py` ran eslint and encountered issues, eslint's default `stylish` formatter emitted multi-line blocks per problem — typically 4-6 lines each — producing ~1000 lines of output for a file with many violations.

## What changed

Added `--format`, `compact` to the eslint command in `do_jslint()` (`dev.py` line 273). The `compact` formatter emits one line per problem in the form `file:line:col: message [rule]`, preserving all issue information while keeping output proportional to the number of problems found.

## Validation

- `uv run --frozen python dev.py lint` — passed
- `uv run --frozen python -m pytest dev_test.py::test_find_files -p no:timeout` — 1 passed (only test relevant to `dev.py`; full suite conftest.py hits the database during initialization which timed out in this environment even after MariaDB was started)

Fixes #12584

---

[Open workspace in Conductor](https://app.conductor.build/workspace/dc9b4c8b-f55a-476f-a2e6-efa852b117d4)

Issue #12584, opened 2024-05-15
jslint's output is about 1000 lines long if it encounters an issue


bakert:
There's a nice summary at the top that's useful. Can we suppress the verbiage?

PR #15218, opened 2026-09-05
Revert ineffective compact jslint formatter

bakert:
## Why this is being reverted

PR #15060 claimed ESLint's default `stylish` formatter emits four to six lines per problem and that `compact` would materially reduce the roughly 1,000-line failure output reported in #12584. That diagnosis is not borne out by the formatter behavior.

With 100 deliberate lint errors against the repository configuration:

- `stylish`: 105 lines, 8,167 bytes
- `compact`: 103 lines, 17,262 bytes

Compact repeats the absolute path on every diagnostic, so it saves only two lines while producing more than twice as much output.

## What this changes

- Restore ESLint's default formatter.
- Remove the unnecessary `eslint-formatter-compact` development dependency and lockfile entry.
- Reopen #12584 so the actual source of the noisy output can be captured and addressed.

The standalone formatter is a valid ESLint v9 package; this revert is about it not solving the ticket, not package correctness.

## Validation

- Clean `npm ci`: 452 packages, 0 vulnerabilities
- `uv run --frozen python dev.py jslint`: passed
- `uv run --frozen python dev.py lint`: passed
- `git diff --check`: passed

Reverts #15060.
Reopens #12584.

### DIFF
--- a/dev.py
+++ b/dev.py
@@ -307,7 +307,7 @@ def pull_request(argv: list[str]) -> None:
 def do_jslint(fix: bool) -> None:
     print('>>>> Linting javascript')
     files = find_files(file_extension='js', exclude=['shared_web/static/js/tipped.min.js']) + find_files(file_extension='jsx')
-    cmd = [os.path.join('.', 'node_modules', '.bin', 'eslint'), '--format', 'compact']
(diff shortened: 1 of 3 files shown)


bakert:
@mergifyio queue

PR #15220, opened 2026-09-05
Use emoji names only as a metadata fallback

bakert:
## Context

Follow-up to #15063 and #12580.

#12580 asked us not to emit large, colorful emoji in `<title>` and social metadata, without broadly stripping Japanese, Chinese, or other safe Unicode. The initial implementation in #15063 removed every Unicode symbol; its merged follow-up instead converted every emoji to its English name.

That solved the rendering problem, but made ordinary deck titles noisy (`No More Meming 👮👮👮` became `No More Meming police officer police officer police officer`) and also rewrote text-presentation symbols such as `™`, `©`, and `♥`.

## Change

Use a predictable hybrid policy:

- Mixed text: remove emoji (`No More Meming 👮👮👮` → `No More Meming`).
- Emoji-only fields: retain the readable fallback (`🔥🐟` → `fire fish`) so titles never become blank.
- Text-presentation Unicode: preserve it exactly (`A Different Jund™ Pile` stays unchanged).
- Process user-provided names before adding fixed metadata text, so an emoji-only name is still recognized as emoji-only.
- Preserve whitespace exactly when no emoji was removed.

There is no reliable general heuristic for distinguishing a semantic rebus such as `Parish Is the 🔪` from decorative emoji, so this intentionally favors a consistent rule over a growing exception list.

## Verification

- `uv run pytest shared/text_test.py decksite/view_test.py decksite/views/metadata_test.py` (17 passed)
- `uv run --frozen python dev.py mypy` (375 files, clean)
- `uv run --frozen python dev.py lint` (clean)
- Headless Chrome against real local deck pages confirmed mixed titles are cleaned, emoji-only titles retain word fallbacks, and `™` remains unchanged.

### DIFF
--- a/decksite/view.py
+++ b/decksite/view.py
@@ -137,15 +137,17 @@ class View(BaseView):
         return url_for('favicon', rest='-152.png')
 
     def title(self) -> str:
-        if not self.page_title():
+        page_title = self.page_title()
+        if not page_title:
             return 'pennydreadfulmagic.com'
+        page_title = text.replace_emoji_with_text(page_title)
         if get_season_id() == seasons.current_season_num():
             season = ''
         elif get_season_id() == 0:
             season = ' - All Time'
         else:
             season = f' - Season {get_season_id()}'
-        return text.replace_emoji_with_text(f'{self.page_title()}{season} – pennydreadfulmagic.com')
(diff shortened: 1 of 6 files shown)


bakert:
@mergifyio queue

PR #15240, opened 2026-09-06
Lock decksite migrations across workers

### LINKED ISSUE #6478: Decksite migrations aren't locked, and multiple threads run them at once
https://github.com/PennyDreadfulMTG/perf-reports/issues/50813

bakert:
## Summary

- serialize decksite migrations across web workers with the existing MariaDB named-lock primitive
- make waiting workers re-read `db_version` after acquiring the lock, so each patch runs once and workers cannot continue startup against a migration still in progress
- release the lock on both success and failure, with a bounded 10-minute startup wait

Closes #6478.

## Why

The migration runner read the current schema version and applied every later SQL file without synchronization. Two workers starting together could therefore choose the same migration. The linked production report for #6478 shows both workers attempting the locale `ALTER TABLE`, with one failing on a duplicate column.

The existing `interprocess_locked()` helper uses the same underlying MariaDB lock for cards and maintenance work, but deliberately skips duplicate background jobs. Migrations instead need to wait and then re-check the schema; skipping would allow new code to start against the old schema.

## Validation

- fail-before: `uv run --frozen pytest decksite/database_test.py -q` ran the simulated migration twice and failed the new serialization assertion
- pass-after: `uv run --frozen pytest decksite/database_test.py -q` — 1 passed
- `uv run --frozen pytest decksite/database_test.py shared/database_test.py shared/decorators_test.py -q` — 14 passed
- `uv run --frozen ruff check decksite/database.py decksite/database_test.py`
- `uv run --frozen mypy decksite/database.py decksite/database_test.py`
- isolated real-MariaDB deployment probe: two independent Python processes ran the actual `decksite.database.setup()` path against a deliberately slow, non-idempotent migration; both exited 0, only one logged the patch, the schema object existed once, and `db_version` contained one row

## Post-deploy checks

- all decksite workers start and remain healthy
- `SELECT MAX(version), COUNT(*) FROM db_version` is unchanged by this code-only deploy
- `SELECT IS_FREE_LOCK('penny-dreadful-tools:decksite:migrations')` returns 1 after startup
- logs contain no lock timeout, `DatabaseException`, duplicate-schema error, or worker restart loop
- on the next genuine migration, exactly one worker logs `Patching database to vN` and exactly one version-N row is recorded

---

[Open workspace in Conductor](https://app.conductor.build/workspace/b4dfbb98-9c19-4118-9beb-0f37ec2fd2d7)

### DIFF
--- a/decksite/database.py
+++ b/decksite/database.py
@@ -8,6 +8,8 @@ from shared.container import Container
 from shared.database import Database, get_database
 
 TEST_CONTEXT = Container()
+MIGRATION_LOCK = 'penny-dreadful-tools:decksite:migrations'
+MIGRATION_LOCK_TIMEOUT_SECONDS = 10 * 60
 
 def db() -> Database:
     ctx: Any = TEST_CONTEXT  # Fallback context for testing.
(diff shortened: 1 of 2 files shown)


bakert:
@mergifyio queue

Issue #15253, opened 2026-09-06
Performance backlog: remote pickers and search


bakert:
Follow-up work discovered while optimizing `/matchups`:

- [ ] Replace the shared admin person dropdown on `/admin/aliases/`, `/admin/people/notes/`, `/admin/unlink/`, and `/admin/ban/` with a remote picker.
- [ ] Replace repeated archetype dropdowns, especially the roughly 12,000 options generated by `/admin/archetypes/`.
- [ ] Extract/reuse a configurable remote typeahead implementation across matchups, admin pickers, and header search.
- [ ] Replace header search’s generated typeahead file and scheduled job with ranked, server-side capped live search.

The admin pickers are higher priority than the header-search rewrite.

PR #15255, opened 2026-09-06
Paginate matchup deck and match results

bakert:
## Summary

- Render matchup decks through the existing paginated, sortable `DeckTable`.
- Render matchup matches through a read-only mode of the existing paginated, sortable `MatchTable`.
- Replace `GROUP_CONCAT` result materialization with accurate deck/match counts and optional opponent filters on the existing `/api/decks/` and `/api/matches/` APIs.
- Preserve the matchup record calculation and all result links while bounding each table response to 20 or 100 rows.
- Preserve the standard deck API's active-league-run visibility rules.

## Pathological case

All-time Aggro vs Aggro previously tried to server-render 33,458 deck rows and 26,192 match rows. It now returns a roughly 34 KB HTML page, then loads 20 rows per table. The tables report the full totals and can page/sort through them.

The summary remains unchanged at 26,181–26,181–22 (50.0%).

## Verification

- `uv run --frozen python dev.py mypy`
- `npm run build`
- 32 focused tests passed.
- 999 broader nonfunctional tests passed; two unrelated tests require a clean database and encountered archetypes already present in the development snapshot.
- Real Chrome check against the cloud snapshot: both tables rendered 20 rows through `/api/decks/` and `/api/matches/`, match pagination advanced to page 2, deck sorting reloaded successfully, and no JavaScript/request errors occurred.

Follow-up backlog: #15253


bakert:
@mergifyio queue

PR #15256, opened 2026-09-06
Decommission broken locale switching

### LINKED ISSUE #6153: Unrecognized locales cause 500s
For example: `/cards/Electrostatic%20Field/?locale=NA`

 - [ ] Do NOT use querystring for locale, it causes search results to be stored in Google than then switch the locale of the viewer.
 - [ ] Send a Bad Request not an Internal Server Error if you ask to be switched to a locale that does not exist.

bakert:
Closes #6153.

## Summary

- remove the broken language switcher and locale/session selection machinery
- treat legacy `locale` query parameters as inert instead of changing language or erroring
- remove stale translation UI, styles, assets, About-page claims, and Discord help
- retain Flask-Babel/gettext as an English-only compatibility layer

## Validation

- targeted regression test demonstrated three failures before and three passes after
- `uv run pytest logsite/smoke_test.py::test_home shared_web/flask_app_test.py decksite/view_test.py -q` (22 passed)
- `uv run python dev.py lint`
- `uv run python dev.py mypy`
- `uv run python dev.py jslint`
- `npm run build`
- browser-verified `/?locale=fr-fr` renders the normal English page without a language switcher

---

[Open workspace in Conductor](https://app.conductor.build/workspace/e93d0b50-d84d-46b8-b71b-b8de39936e82)


bakert:
@mergifyio queue

bakert:
@mergifyio queue

PR #15277, opened 2026-09-07
Add loading skeletons to live tables

### LINKED ISSUE #15266: Reduce layout shift while live tables load
## Problem

Live tables initially render an empty server-side mount point, then a one-line React loading indicator, and finally the populated table. The shared `Table` component does not reserve space for its loaded state, so content below the table moves substantially when the API response arrives.

This is site-wide rather than specific to `/matchups`. A delayed browser check at a 1440×900 viewport measured:

- `/decks/`: roughly 20 px while loading, then roughly 681 px with 20 rows.
- `/matchups/`: two roughly 20 px loading indicators, then roughly 691 px and 661 px tables.

The relevant shared implementation is `shared_web/static/js/table.jsx`; all live-table pages mount into empty `div.decktable`, `div.matchtable`, etc. elements.

## Desired outcome

Reduce loading-time layout shift consistently across all live tables. Use a shared treatment rather than a matchup-only override—for example a table-shaped skeleton or a sensible reserved-height/loading container that remains responsive and does not leave excessive empty space for short/empty results.

## Acceptance checks

- `/decks/`, `/people/{id}/`, and populated `/matchups/` do not visibly jump when their initial table requests complete.
- Empty and small result sets do not retain a large blank area after loading.
- Pagination and sorting retain their current loading behavior and remain accessible.
- The behavior works at desktop and narrow/mobile widths.

bakert:
## Summary

- render each live table's real headers and a page-sized skeleton during its initial API request
- reserve row and pagination space using the table's configured page size, conditional columns, and season-icon layout
- keep the loading state accessible and disable its animation for reduced-motion users
- make the browser-test helper wait for real rows instead of accepting skeleton rows, with a regression test that delays the decks API

## Verification

- `npm run build`
- `uv run --frozen mypy decksite/browser_test.py`
- `git diff --check`
- manually delayed API responses in Chrome for `/decks/`, `/people/carmandor/`, and `/matchups/` at desktop and mobile widths
- verified the 20-row `/decks/` skeleton is within 0.5 px of the loaded table height and does not cause document-level horizontal overflow on mobile

The focused Playwright regression test cannot complete in this cloud workspace because the local MariaDB user cannot create the test database (`decksite_test_seeded`). Its setup reaches that existing infrastructure restriction before the test runs.

Closes #15266.

### DIFF
--- a/decksite/browser_test.py
+++ b/decksite/browser_test.py
@@ -93,10 +93,12 @@ def wait_for_live_tables(page: 'Page') -> list[str]:
             root = page.locator(f'div.{css_class}').nth(i)
             rows = root.locator('.metagame-grid > *' if css_class == 'metagamegrid' else 'table tbody tr')
             try:
+                if css_class != 'metagamegrid':
+                    expect(root.locator('.table-skeleton')).to_have_count(0, timeout=LOAD_TIMEOUT)
                 expect(rows.first).to_be_attached(timeout=LOAD_TIMEOUT)
             except AssertionError:
                 error = root.locator('.message.error')
-                detail = error.first.text_content() if error.count() else ('still loading' if root.locator('.loading').count() else 'no rows')
+                detail = error.first.text_content() if error.count() else ('still loading' if root.locator('.loading, .table-skeleton').count() else 'no rows')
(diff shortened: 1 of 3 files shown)


bakert:
@mergifyio queue

PR #15284, opened 2026-09-07
Fix archetype picker assignment

bakert:
## Summary
- keep value pickers scoped to their paired hidden field
- prevent archetype selection from overwriting the deck ID
- add browser regression coverage for adjacent hidden form values

## Verification
- 34 admin controller tests pass
- targeted Chromium regression test passes
- JavaScript and Python lint pass
- production JavaScript build passes

---

[Open workspace in Conductor](https://app.conductor.build/workspace/408e8a0c-daad-4ad8-a384-3e59c51c1866)


bakert:
@mergifyio queue

